PR #56, opened 2020-08-11
fix(Gradle): Antlr sources dir incorrect

### LINKED ISSUE #55: Antlr sources dir incorrect in Gradle
It is set to $buildDir/gensrc/org/python/antlr whereas it should be $buildDir/gensrc. This gives errors when importing the project in Eclipse

SchwingSK:
Fixes #55


SchwingSK:
Unless I'm mistaken, changing Gradle config shouldn't have triggered an error on the ant-based Travis CI. Is this a false positive?

jeff5:
Yes, the tests are flakey, and I'm sure that failure on Windows is unrelated.

It works as written (javac doesn't care), but in yours with separate variables the meaning is clearer and maybe IDEs will be happier.

This seems small enough not to bother with the Python CLA.

Issue #60, opened 2020-08-19
sendall of more than 64k not working


gbach:
https://github.com/jython/jython/blob/0a58cc26566d2b2334e80b2b3f2f42f6c738db2d/Lib/_socket.py#L1221

to fix:

```python 
def sendall(self, data, flags=0):
        chunk_size = 8192
        length = len(data)
        data_view = memoryview(data)
        idx = 0
        while idx < length:
            bytes_sent = self.send(data_view[idx:idx + chunk_size], flags=flags)
            idx += bytes_sent
```

behackett:
Related - https://bugs.jython.org/issue2618

PR #80, opened 2021-01-22
FIX: Send all > 64k not working

gbach:
See https://github.com/jython/jython/issues/60

### DIFF
--- a/Lib/_socket.py
+++ b/Lib/_socket.py
@@ -1218,8 +1218,15 @@ class _realsocket(object):
 
         return len(sent_data)
 
-    sendall = send   # FIXME see note above!
-
+    def sendall(self, data, flags=0):
+            chunk_size = 8192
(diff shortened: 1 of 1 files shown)


jeff5:
Removed from milestone 2.7.3 because #185 is expected to resolve (within a somewhat broader in scope).

jeff5:
@gbach : I've merged similar code as part of #185, so we can close this. Thanks however for your interest in solving it and keeping this on the to-do list, which has led to tidying up several loose ends in this bit of `_socket.py`.

The use of a `memoryview` is good, and it seemed to improve performance, used consistently, in `send()` too.

You have to remember to dispose of them because they lock the underlying object. Even a slice of a  `memoryview` is a  `memoryview`, which I forgot in my first change set. `with` is a good idiom for this.

PR #81, opened 2021-01-24
Support "from __future__ import print_function" in interactive interpreters (fixes issue 2007)

shoaniki:
In CPython, `from __future__ import print_function` works both in scripts and interactive interpreters; in the latter it can be enabled at any time and will apply to the parsing of all subsequent inputs.  The same is true of `unicode_literals`.

Jython implements this correctly in scripts, but it has never worked in interactive interpreters.  This was reported early in Jython 2.7 development (https://bugs.jython.org/issue2007), but apparently today is the first time it's been sufficiently annoying for someone to dig into Jython guts and fix it ;)

The cause is that each input is handled by a new PythonParser, and PythonParser currently only enables future features if it sees the import statement itself.  Fortunately the feature flags are also recorded in PythonInterpreter.cflags, and this is already passed to the ParserFacade that creates the parser instances, so it is straightforward to add the missing step and initialize the parser correctly.


jeff5:
@shoaniki : It's taken a while to get to this, but I really appreciate you digging into the guts to fix this. Let's see if we can get it in the next release.

I'll ask you to sign the [Python Contributor Agreement](https://www.python.org/psf/contrib/contrib-form/) before we can merge this, but meanwhile, I'll pull it down for review.

src/org/python/core/CompilerFlags.java
```diff
@@ -61,7 +61,8 @@ public boolean isFlagSet(CodeFlag flag) {
     public String toString() {
         return String.format(
                 "CompilerFlags[division=%s nested_scopes=%s generators=%s "
-                        + "with_statement=%s absolute_import=%s only_ast=%s "
+                        + "with_statement=%s absolute_import=%s "
+                        + "print_function=%s unicode_literals=%s only_ast=%s "
```
jeff5, line 65: I think you have written what the author intended before his attention wandered. :)

jeff5 (changes requested):
This is good! I traced through the un-patched version and it's really difficult to spot where the necessary transfer of options **isn't** happening, so well done for finding what's missing.

No actual code changes are needed.

I'd like to merge this on condition only that you sign the [Python Contributor Agreement](https://www.python.org/psf/contrib/contrib-form/). Let us know it's happened on the PR -- the PSF process is changing we're not properly joined up yet.

jeff5:
I'll push a few change sets onto the PR to bring it up to the current head, so the squash-merge works, and add you to the acknowledgments file. Is your GitHub name @shoaniki all you want to appear?

Unfortunately, this will make a mess of your clone of our repo. (The clean way was to create a new branch and PR from there. We use essentially [the same process as for CPython](https://devguide.python.org/pullrequest/) with the C-specific things adjusted.)

shoaniki:
Whoops, I completely forgot about this too, until just now when I was wondering why print_function wasn't working and came looking to see if anyone had run into the problem before 😅

I have now signed the CLA.

Re acknowledgements, yes, just my Github handle is fine thank you.

jeff5:
That was close. I built a 2.7.3b1 last night, but will call it a dry-run.

I assumed GitHub would ping you about the "change required" -- just the CLA, in practice -- and was disappointed not to be able to include this, so thanks for signing, @shoaniki.

You'll see some more pushes onto your repo to deconflict the merge.

jeff5 (approved):

PR #88, opened 2021-03-17
Travis CI: Add comment linking to homebrew-openjdk

cclauss:
https://travis-ci.org/github/jython/jython/pull_requests should be migrated to https://travis-ci.com instead of .org


jeff5:
Looks good to me.

On the migration question, I've looked at that a couple of times and wasn't entirely sure what it would do if I pushed the button. I should probably just take it on trust.

PR #89, opened 2021-03-17
GitHub Action to lint Python code

cclauss:
Run [codespell](https://github.com/codespell-project/codespell) to find typos.
Run [flake8](https://flake8.pycqa.org) to find Python 3 SyntaxErrors in `./Demo` and `./tests`

Output: https://github.com/cclauss/jython-2/actions

https://flake8.pycqa.org/en/latest/user/error-codes.html

On the flake8 test selection, this PR does _not_ focus on "_style violations_" (the majority of flake8 error codes that [__psf/black__](https://github.com/psf/black) can autocorrect).  Instead, these tests are focus on runtime safety and correctness:
* E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST).  Often these issues are a sign of unused code or code that has not been ported to Python 3.  These would be compile-time errors in a compiled language but in a dynamic language like Python, they result in the script halting/crashing on the user.
* F63 tests are usually about the confusion between identity and equality in Python.  Use ==/!= to compare str, bytes, and int literals is the classic case.  These are areas where __a == b__ is True but __a is b__ is False (or vice versa).  Python >= 3.8 will raise SyntaxWarnings on these instances.
* F7 tests logic errors and syntax errors in type hints
* F82 tests are almost always _undefined names_ which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3.  These also would be compile-time errors in a compiled language but in Python, a __NameError__ is raised which will halt/crash the script on the user.

### DIFF
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -1,0 +1,20 @@
+name: lint_python
+on: [pull_request, push]
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install bandit black codespell flake8 isort mypy pytest pyupgrade safety
+      - run: bandit -r . || true
+      - run: black --check . || true
+      - run: codespell --quiet-level=2 || true  # --ignore-words-list="" --skip=""
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+      - run: isort --check-only --profile black . || true
+      - run: pip install -r requirements.txt || true
+      - run: mypy --ignore-missing-imports . || true
+      - run: pytest . || true
+      - run: pytest --doctest-modules . || true
+      - run: shopt -s globstar && pyupgrade --py36-plus **/*.py || true
+      - run: safety check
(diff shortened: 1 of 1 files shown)


jeff5:
Thanks for presenting this. My skill with the build and CI is rudimentary.

There is still a lot of legacy from Jython 2 lying around. It is not referenced in the build: it's just there awaiting conversion or deletion. This may take a while. I wouldn't want it an action to nag us about it being Python 2 for all that time.

It would be good to convert the demo code (when there is anything to demonstrate). Stuff in the tests folder is probably junk.

I have some Python for code-generation (in the pipeline), and that ought to be good Python 3 (but isn't). An action like this would run on a PR, wouldn't it?

cclauss:
The Python 3 linting currently only looks at `./Demo` ~and `./tests`~ -- Adjustments are always possible.

> An action like this would run on a PR, wouldn't it?

I could run on each PR or could run like a cron job (once a day, week, month, year).

jeff5:
Hi Christian (@cclauss). I think this is a good thing in principle, but the `main` branch (Jython 3) holds over a lot of Jython 2 code (and Python 2 code) that is simply waiting to be incorporated or deleted.

Only the sub-project `build-tools` and `core` contain live code.  Of these only `build-tools` contains Python. I foresee some Python appearing in `core` in the near future (in `test`), but I'm having to focus on Jython 2 for a time. :(

This will all be Python 3.8 code for as long as that's what we're targeting.

jeff5:
I still think this is a good thing. I'd use it first in my local environment (where I don't always use an IDE that is fussy about style).

Later it should perhaps be part of the build for PRs.

Issue #100, opened 2021-06-24
Arguments not being passed to vararg methods correctly


tpoliaw:
If a Java method such as

    public static void foo(Object... args) {}

is called from Jython as

    foo(1, 2, [3,4])

The call should be valid and the `args`that gets passed to the Java method should be `[1, 2, [3, 4]]`.

The current method checks the final arg to check if it's a sequence and assumes the varargs are already being passed as a list so the call fails. A similar call without the list (eg `foo(1, 2, 3)`) works as expected.

A similar bug (but the other way around) can be see by calling a Java method such as

    static void foo(int[] a, int... b) {}

as 

    foo([1,2,3])

which should be valid, but needs an additional empty list added to fit the vararg signature.

tpoliaw:
I think this is fixed by #101. Thanks for reviewing it.

jeff5:
> I think this is fixed by #101. Thanks for reviewing it.

Right, I should've mentioned this one in the consolidate commit message, so it would get closed. Thanks to you for solving it.

PR #101, opened 2021-06-24
Improve boxing of arguments to vararg methods

tpoliaw:
Potential fix for #100.

Only check final argument for a sequence if the correct number of arguments have been passed.


tests/java/org/python/core/ReflectedArgsTest.java
```diff
@@ -0,0 +1,53 @@
+package org.python.core;
+
+import junit.framework.TestCase;
+
+/**
+ * Tests for ReflectedArgs.
+ */
+public class ReflectedArgsTest extends TestCase {
+	static class Demo {
+		public static void foo(Object... args) {}
```
Stewori: Please use four spaces for indentation rather than tabs. (I am actually in favor of tabs in general, but one should stick to the project's style in this case)

tpoliaw: Cheers. I matched the spaces in the other file but didn't think to check the new file.

tpoliaw:
Any chance of getting some feedback on this? It would be good to get this fixed without having to maintain the patched version locally.
Happy to rework the PR if required.

jeff5:
@tpoliaw  Sorry, I've been having too much fun with Jython 3, but I'm having a push on Jython 2 issues now. I hope to accept one soon. It looks well-reasoned, and to have Java tests is reassuring.

Single-stepping through some of those will improve my understanding of argument matching. Then my opinion will be worth something!

I don't understand the "force push" on this PR. When I pull it to a branch on my repo I only see one change set (`9fb81f4`). However, if the difference is only white space I won't worry.

tpoliaw:
The force push was to update the commit with the whitespace changes  @Stewori requested rather than adding a new reformatting commit. The diff for before/after the update is [here](https://github.com/jython/jython/compare/c373c358f387714773b49c12e1d4cc08ccb8418f..9fb81f49a40e444fd300bc3e699fd072259c3fbc)

jeff5:
That's what I guessed.

This looks good functionally, thanks. I'm merging in the tip of master for test, and will push that to your PR branch. Then there are some non-code changes (e.g. line length, and making your helpful commit message a comment) but I'll make and push those rather than ask for them.

I'll squash-merge at the end which should make a single change set in which all the extra change will have evaporated.

At least, I think that will work.

jeff5:
@tpoliaw : I set out to give each test a Javadoc comment based on the comment in the body, but on close inspection, a couple of the the tests didn't actually do what they claim, so I adjusted the code to match the comment. That led me to want stronger tests of what actually comes back from `match` in  `callData` , and to have both `bar(int i, int[] j, int... rest)` and `baz(Integer i, Integer[] j, Integer... rest)`.

So the test is a lot longer now, and I understand what the code actually does. :)

Have I got this right and do they now cover the range of your addition to `ReflectedArgs`?

jeff5:
And in a further confirmation, if `Demo` is made `public`, this all delivers the arguments one would hope:
```
Jython 2.7.3a1-DEV (heads/pr-101-merge:7b138dcaa, Mar 20 2022, 11:10:07)
[Java HotSpot(TM) 64-Bit Server VM (Oracle Corporation)] on java1.8.0_321
>>> from org.python.core.ReflectedArgsTest import Demo
>>> Demo.foo(3, 4, ["bar"])
>>> Demo.foo([1,2,3])
>>> Demo.foo(1,2,3)
>>> Demo.foo("foo", "bar")
>>> Demo.bar(1, [2, 3])
>>> Demo.bar(1, [2, 3], 4, 5)
>>> Demo.bar(1, [2, 3], [4, 5])
>>> Demo.baz(1, [2, 3])
>>> Demo.baz(1, [2, 3], 4, 5)
>>> Demo.baz(1, [2, 3], [4, 5])
```

tests/java/org/python/core/ReflectedArgsTest.java
```diff
@@ -1,53 +1,167 @@
...
+        PyObject[] ints = {Py.newInteger(1), Py.newInteger(2), Py.newInteger(3)};
+        PyObject[] pyArgs = {new PyList(ints)};
+        ReflectedCallData callData = new ReflectedCallData();
+        assertTrue(FOO_SIGNATURE.matches(null, pyArgs, Py.NoKeywords, callData));
+        fooCheck(callData.args, 1, 2, 3);
     }
 
+    /**
+     * Calling {@code foo(Object... args)} as Python {@code foo("foo", "bar")} calls Java
+     * {@code foo("foo", "bar")}.
```
tpoliaw, line 158: should this be `calls Java {@code foo(new String[] {"foo", "bar"})}`?

tests/java/org/python/core/ReflectedArgsTest.java
```diff
@@ -1,53 +1,167 @@
...
-     * eg foo(1, [1, 2, 3]) should be valid for a call to
-     * void foo(int i, int[] j, int... rest)
-     * but should be boxed as foo(1, [1, 2, 3], [])
+     * eg foo(1, [1, 2, 3]) should be valid for a call to void foo(int i, int[] j, int... rest) but
+     * should be boxed as foo(1, [1, 2, 3], [])
      */
 
+    /**
+     * Calling {@code foo(Object... args)} as Python {@code foo(3, 4, ["bar"])} calls Java
+     * {@code foo(3, 4, strlist})} where {@code strlist} is the representation of Python
```
tpoliaw: looks like a mismatched `}` after `strlist`

tpoliaw (commented):
Thanks for looking at this and improving the docs for the tests. I think everything looks ok to me. 
A couple of nitpicking comments on the javadoc but nothing that should block the PR.

tests/java/org/python/core/ReflectedArgsTest.java
jeff5: I think I wrote `{@code foo(new Object[] {"foo", "bar"})}` first, but as that is quite hard to read and `foo` is a varargs in Java, I think what I have is ok.

tests/java/org/python/core/ReflectedArgsTest.java
jeff5: And here I wrote `{@code foo(new Object[] {3, 4, strlist})}` first (or something worse) then having reconsidered I didn't delete enough of it.

jeff5:
Sigh ... it fails irrelevantly (test_select), then succeeds on the rebound.

PR #102, opened 2021-06-25
Fix CRC32#update calls in zlib.py

hansiang93:
the update method within CRC32.java does not contain a signature that only accepts a single String input, it only accepts either (byte[], int, int) or (ByteBuffer), to which the byte[] input was an easier choice here

### DIFF
--- a/Lib/zlib.py
+++ b/Lib/zlib.py
@@ -19,7 +19,6 @@ import jarray
 import struct
 import sys
 from cStringIO import StringIO
-from org.python.core.util import StringUtil
 
 from java.lang import Long, String, System
 from java.util.zip import Adler32, CRC32, Deflater, Inflater, DataFormatException
@@ -149,7 +148,7 @@ class compressobj(object):
         self.deflater.setInput(string, 0, len(string))
         deflated = _get_deflate_data(self.deflater)
         self._size += len(string)
-        self._crc32.update(StringUtil.toBytes(string), 0, len(string))
+        self._crc32.update(string, 0, len(string))
         if self._gzip:
             return self.GZIP_HEADER + deflated
         else:
@@ -227,7 +226,7 @@ class decompressobj(object):
 
         self.inflater.setInput(string)
         inflated = _get_inflate_data(self.inflater, max_length)
-        self._crc32.update(StringUtil.toBytes(inflated), 0, len(inflated))
+        self._crc32.update(inflated, 0, len(inflated))
 
         r = self.inflater.getRemaining()
         if r:
--- a/Lib/zlib.py
+++ b/Lib/zlib.py
@@ -19,6 +19,7 @@ import jarray
 import struct
 import sys
 from cStringIO import StringIO
+from org.python.core.util import StringUtil
 
 from java.lang import Long, String, System
 from java.util.zip import Adler32, CRC32, Deflater, Inflater, DataFormatException
@@ -148,7 +149,7 @@ class compressobj(object):
         self.deflater.setInput(string, 0, len(string))
         deflated = _get_deflate_data(self.deflater)
         self._size += len(string)
-        self._crc32.update(string)
+        self._crc32.update(StringUtil.toBytes(string), 0, len(string))
         if self._gzip:
             return self.GZIP_HEADER + deflated
         else:
@@ -226,7 +227,7 @@ class decompressobj(object):
 
         self.inflater.setInput(string)
         inflated = _get_inflate_data(self.inflater, max_length)
-        self._crc32.update(inflated)
+        self._crc32.update(StringUtil.toBytes(inflated), 0, len(inflated))
 
         r = self.inflater.getRemaining()
         if r:
--- a/NEWS
+++ b/NEWS
@@ -11,6 +11,7 @@ For more details, three sources are available according to type:
 
 Jython 2.7.3a1
   Bugs fixed
+    - [ GH-102 ] Fix CRC32.update calls in zlib.py
     - [ GH-35 ] Travis CI on JDKs 8, 11, 13 and add Windows to OSes
     - [ GH-50 ] (First) migration from Mercurial corrupted project history
     - [ GH-27 ] -Q new always fails


jeff5:
Hi @hansiang93:  I've been looking at this today. Thanks for pointing it out. You have definitely identified how this is going wrong.

I wonder why this is not revealed by the failure of any test? This is a gap in coverage that should be fixed in `test_zlib_jy`.

The solution isn't quite right, however. In the case where `string` is a `str` (`PyString` internally), it is not necessary explicitly to invoke `StringUtil.toBytes`. The process of calling a Java method tries to find a conversion for each argument by asking `PyObject.__tojava__`. `PyString.__tojava__` knows how to satisfy the request for a `byte[]` argument. In fact, it uses `StringUtil.toBytes`.

If you set a debugger breakpoint in `StringUtil.toBytes` you will see this happen during the call to `Deflater.deflate(byte[] b, int off, int len)` a couple of lines up from your change.

So passing `string` to `self._crc32.update` in the same way should do the trick. This seems to work. Did you try it and it failed in your case?

jeff5:
I wrote above "In the case where `string` is a `str`" because `string` is produced by `_to_input()` which which may return several different types of thing: `str`, `array.array`, `buffer`, `memoryview`.

Someone (@jimbaker it appears) has taken the trouble to add these last two cases to `_to_input()`, but they won't work because when I implemented them I wouldn't have been able to explain `__tojava__` as I have, so they don't know how to match themselves to arguments. :( `bytearray` should perhaps be in the list, and my excuse would be the same.

There is a semantic problem with `__tojava__` where it makes a copy, which is right for immutable Python types, but will not behave as may have been expected where the Java method is supposed to modify the argument. It's ok here. Going further, it would be really cool to be able to satisfy a Java ByteBuffer argument with any Python buffer type.

I mention all this in a separate comment because I think it goes well beyond fixing the bug, but it would be good to study as a separate change.

Stewori:
**Without having had a look at the code in this PR** I would like to point out that we once fixed a ``String`` versus ``PyString`` issue by adjusting the API to work with ``CharSequence`` instead, which I consider a rather elegant solution. For that I took care (once upon a time) that ``PyString`` implements ``CharSequence`` properly. As a side effect such a solution also integrates seamlessly with ``StringBuilder`` and ``StringBuffer``. I acknowledge that it is not always practical to operate on top of ``CharSequence`` though. Maybe that idea can be of some use here...

jeff5:
@Stewori : That is certainly a helpful reminder in relation to `PyUnicode` in Jython 3. Here the Java methods are those of `java.util.zip.*` which all want a `byte[]` or `ByteBuffer` argument.

jeff5:
@hansiang93 : Looking at this harder, I realise don't understand what problem is being solved.

> the update method within CRC32.java does not contain a signature that only accepts a single String input, it only accepts either (byte[], int, int) or (ByteBuffer), to which the byte[] input was an easier choice here

There is a method [`CRC32.update(byte[])`](https://docs.oracle.com/javase/7/docs/api/java/util/zip/CRC32.html#update(byte[])) and a Python `str` can adapt itself to `byte[]` during argument processing. If a problem occurs at this point, it is because something other than `str` is being supplied to compress.

This is possible because _to_input incorrectly lets though types that don't support conversion to `byte[]`. They probably should. But I don't see how the offered solution fixes that.

The question then is what error were you observing with the unmodified Jython? What kind of object were you supplying to the compressor?

hansiang93:
The actual issue here isn't that str couldn't be converted into a byte[]. It can, but the actual issue here is a mismatched method signature as an both ByteBuffer and byte[] matches "Object".

The main issue is that with a "pyString object", this matches both a ByteBuffer and a byte[]. depending on the JVM, either one of these 2 signatures could be matched at runtime, and a runtime cohersion exception would happen if the jvm decided to use a ByteBuffer signature instead of the byte[] method.

Here's the error stack trace I received:


  File "/home/...../mysecretfile.py", line ..., in ....
    lines = f.readlines()
  File "/home/..../Lib/gzip.py", line 473, in readline
    c = self.read(readsize)
  File "/home/..../gzip.py", line 268, in read
    self._read(readsize)
  File "/home/..../gzip.py", line 268, in read
    self._read(readsize)
  File "/home/..../gzip.py", line 319, in _read
    uncompress = self.decompress.decompress(buf)
  File "/home/..../zlib.py", line 229, in decompress
    self._crc32.update(inflated)
TypeError: update(): 1st arg can't be coerced to java.nio.ByteBuffer, int

I traced this to the updated Zlib.py: https://github.com/jython/jython/commit/a2f0e48b930681d2fe1011329c8c9082b92a88b0#diff-811e4932de03839c2b93b67ba1b9da30014cdc839ca6eceee338697bf0e20ab0R229 

from cStringIO import StringIO

....
        inflated = _get_inflate_data(self.inflater, max_length)
        self._crc32.update(inflated)

.....

def _get_inflate_data(inflater, max_length=0):
    buf = jarray.zeros(1024, 'b')
    s = StringIO()
    ....
    return s.read()

and in cStringIO.java, a PyString is returned: https://github.com/jython/jython/blob/0a58cc26566d2b2334e80b2b3f2f42f6c738db2d/src/org/python/modules/cStringIO.java#L240 

        public PyString read() {
            return read(-1);
        }

So i had concluded that the JVM was unable to parse PyString into byte[] smartly and decided to match a ByteBuffer instead.

jeff5:
> The actual issue here isn't that str couldn't be converted into a byte[]. It can, but the actual issue here is a mismatched method signature as an both ByteBuffer and byte[] matches "Object".
>
> The main issue is that with a "pyString object", this matches both a ByteBuffer and a byte[]. depending on the JVM, either one of these 2 signatures could be matched at runtime, and a runtime cohersion exception would happen if the jvm decided to use a ByteBuffer signature instead of the byte[] method.

It's not the JVM's choice. The decisive match occurs here:
https://github.com/jython/jython/blob/7a073156587066e1e39121156b70c5a80e6b4756/src/org/python/core/ReflectedArgs.java#L145-L157
You see it asks the object passed as argument whether it can turn itself into an instance of the class that appears as a parameter. A `PyString` is able to turn itself into a `byte[]` but not a `ByteBuffer`.

The stack trace you show is critical to our understanding, thanks. The fact that it ends `can't be coerced to java.nio.ByteBuffer, int` tells me it didn't even try `byte[]`. If I abuse `CRC32.update` on my machine I get:
```
>>> c = CRC32()
>>> c.update(1.0)
TypeError: update(): 1st arg can't be coerced to byte[], java.nio.ByteBuffer, int
```
 
The light may have dawned. [`update(byte[] b)`](https://docs.oracle.com/javase/8/docs/api/java/util/zip/CRC32.html#update-byte:A-), present in Java 8 has disappeared from [CRC32](https://docs.oracle.com/javase/9/docs/api/java/util/zip/CRC32.html) in Java 9. Now we can usually expect a decent deprecation period from Java, but I see that it has a default implementation in the interface [`Checksum`](https://docs.oracle.com/javase/9/docs/api/java/util/zip/Checksum.html#update-byte:A-).

To my relief, as I don't think we planned it so, Jython deals correctly with this `default` inheritance. So in fact the method we need is still there in my Jython:
```
Jython 2.7.2 (v2.7.2:925a3cc3b49d, Mar 21 2020, 10:03:58)
[Java HotSpot(TM) 64-Bit Server VM (Oracle Corporation)] on java11.0.9
>>> from java.util.zip import CRC32, Checksum
>>> for sig in CRC32.update.argslist: print sig.method
...
public void java.util.zip.CRC32.update(int)
public default void java.util.zip.Checksum.update(byte[])
public void java.util.zip.CRC32.update(java.nio.ByteBuffer)
public void java.util.zip.CRC32.update(byte[],int,int)
>>> for sig in Checksum.update.argslist: print sig.method
...
public abstract void java.util.zip.Checksum.update(int)
public default void java.util.zip.Checksum.update(byte[])
public default void java.util.zip.Checksum.update(java.nio.ByteBuffer)
public abstract void java.util.zip.Checksum.update(byte[],int,int)
```
I'm still unable to reproduce the problem on available JVMs. But do you perhaps have one where either this change (to implement `CRC32.update(byte[])` in `Checksum.update(byte[])`) has been bungled, or one where Jython does not pick up `default` definitions? What does the sequence above give you? Which JVM and OS?

hansiang93:
Looks like mine is inherently different:

```
>>> 
>>> from java.util.zip import CRC32, Checksum
>>> for sig in CRC32.update.argslist: print sig.method
...
public void java.util.zip.CRC32.update(int)
public void java.util.zip.CRC32.update(java.nio.ByteBuffer)
private static native int java.util.zip.CRC32.update(int,int)
public void java.util.zip.CRC32.update(byte[],int,int)
>>> for sig in Checksum.update.argslist: print sig.method
... 
public abstract void java.util.zip.Checksum.update(int)
public default void java.util.zip.Checksum.update(byte[])
public default void java.util.zip.Checksum.update(java.nio.ByteBuffer)
public abstract void java.util.zip.Checksum.update(byte[],int,int)
>>> 
```

And i am using the following java version:
```
java -version
openjdk version "11.0.12" 2021-07-20 LTS
OpenJDK Runtime Environment Zulu11.50+19-CA (build 11.0.12+7-LTS)
OpenJDK 64-Bit Server VM Zulu11.50+19-CA (build 11.0.12+7-LTS, mixed mode)
```

It's odd that the method still exists for your jython. just to be sure, i'm using a CentOS system with java 11, both running on Zulu or OpenJDK's java 11 runtime causes this.

jeff5:
> It's odd that the method still exists for your jython. 

No, it's odd that *doesn't* exist on yours. :)

However, we now have another clue, which is the presence of a private method. This tells me you are disrespecting Java accessibility. Now I can reproduce it:
```
PS gh-102> jython -J"-Dpython.security.respectJavaAccessibility=false"
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.python.core.PyJavaType (file:/C:/Jython/2.7.2/jython.jar) to method java.lang.Object.finalize()
WARNING: Please consider reporting this to the maintainers of org.python.core.PyJavaType
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
Jython 2.7.2 (v2.7.2:925a3cc3b49d, Mar 21 2020, 10:03:58)
[Java HotSpot(TM) 64-Bit Server VM (Oracle Corporation)] on java11.0.9
>>> from java.util.zip import CRC32, Checksum
>>> for sig in CRC32.update.argslist: print sig.method
...
public void java.util.zip.CRC32.update(int)
public void java.util.zip.CRC32.update(java.nio.ByteBuffer)
private static native int java.util.zip.CRC32.update(int,int)
public void java.util.zip.CRC32.update(byte[],int,int)
>>> for sig in Checksum.update.argslist: print sig.method
...
public abstract void java.util.zip.Checksum.update(int)
public default void java.util.zip.Checksum.update(byte[])
public default void java.util.zip.Checksum.update(java.nio.ByteBuffer)
public abstract void java.util.zip.Checksum.update(byte[],int,int)
>>> import zlib
>>> comp = zlib.compressobj()
>>> data = comp.compress("hello")
  File "C:\Jython\2.7.2\Lib\zlib.py", line 151, in compress
    self._crc32.update(string)
TypeError: update(): 1st arg can't be coerced to java.nio.ByteBuffer, int
```

So the actual bug is that default implementations are not inherited if `python.security.respectJavaAccessibility=false`. That's very odd.

Our attention has been drawn to the fact we don't deal with `default` methods properly in https://bugs.jython.org/issue2866, but I've been ignoring it because it sounded hard. (As it's not exactly a "new feature", that's a little neglectful.) The fact that it works sometimes gives hope it is less hard, and that the question of when it works is entangled with `python.security.respectJavaAccessibility` narrows it to just a few places.

Now I understand what problem this PR actually solves: a bug that showed up because of a change between Java 8 and 9 that Java users would probably not notice. It treats one place that our accidental support for default implementations falls short. That's ok I think as a work-around.

A PR that showed us how to support default implementations intentionally would be a much more significant outcome from our investigation if you're able to propose it.

Lib/zlib.py
```diff
@@ -19,6 +19,7 @@
 import struct
 import sys
 from cStringIO import StringIO
+from org.python.core.util import StringUtil
```
jeff5: It is not necessary to use `StringUtil.toBytes` explicitly. Now we understand why it is necessary to call the 3-argument form of `update` we can be assured that Jython will convert `str` to `byte[]` by itself, and with less copying.

Lib/zlib.py
```diff
@@ -148,7 +149,7 @@ def compress(self, string):
         self.deflater.setInput(string, 0, len(string))
         deflated = _get_deflate_data(self.deflater)
         self._size += len(string)
-        self._crc32.update(string)
+        self._crc32.update(StringUtil.toBytes(string), 0, len(string))
```
jeff5: Just use the 3-argument form.

Lib/zlib.py
```diff
@@ -226,7 +227,7 @@ def decompress(self, string, max_length=0):
 
         self.inflater.setInput(string)
         inflated = _get_inflate_data(self.inflater, max_length)
-        self._crc32.update(inflated)
+        self._crc32.update(StringUtil.toBytes(inflated), 0, len(inflated))
```
jeff5: ... and here.

jeff5 (changes requested):
Good with that simplification. (Lots of other interesting issues in the discussion, for another time.)

jeff5:
To summarise ...  `test_zlib` and `test_zlib_jy` both pass on Java 8 and Java 11 when Jython is allowed to respect the Java accessibility rules.

They fail on Java 11 (and not on Java 8) when `python.security.respectJavaAccessibility=false` because the single argument calls to `CRC32.update` in `zlib.py` get identified with a method introduced after Java 8.

To reproduce in the dev environment, with a Java 11 JDK, use the command:
```
dist\bin\jython  -J"-Dpython.security.respectJavaAccessibility=false" -m test.test_zlib_jy
```
and expect the errors:
```
...
  File "C:\Users\Jeff\Documents\Eclipse-Q\Jython-2\dist\Lib\zlib.py", line 229, in decompress
    self._crc32.update(inflated)
TypeError: update(): 1st arg can't be coerced to java.nio.ByteBuffer, int
...
  File "C:\Users\Jeff\Documents\Eclipse-Q\Jython-2\dist\Lib\zlib.py", line 151, in compress
    self._crc32.update(string)
TypeError: update(): 1st arg can't be coerced to java.nio.ByteBuffer, int
```

The root cause is that when `python.security.respectJavaAccessibility=false` we change our strategy for finding and exposing methods. (I do not understand why.) A work-around (this PR) for the particular case of `CRC32.update` in `zlib.py` is to call it with the 3-argument signature, which is unambiguous.

jeff5:
I've pushed the changes I recommended and merged. I think this can close it (unless tests fail for a related reason).

jeff5:
Looks good. Thanks for raising, @hansiang93.

Issue #93, opened 2021-04-08
Permissions problems with compiled source files ($py.class)


joeshannon:
Jython 2.7.2 includes a fix for https://bugs.jython.org/issue2044 which explicitly sets the permissions on the compiled class files to  u=rw/600 (unix permissions) - this  is set in `org.python.core.util.FileUtil.makePrivateRW`.

Our application is deployed to a network file share where it is run by multiple users (some with write access). With Jython 2.7.2 we are getting failures:
```
IOError: [Errno 2] File not found - /some/path/my_module$py.class (Permission denied)

	at org.python.core.Py.fromIOException(Py.java:214)
	at org.python.core.Py.IOError(Py.java:183)
	at org.python.core.imp.makeStream(imp.java:218)
	at org.python.core.imp.loadFromSource(imp.java:929)
	at org.python.core.imp.find_module(imp.java:769)
...
```
This happens when the second user runs the application, after the class file has been written with the restricted permissions.

Are there any known workarounds for this? We have been setting `dont_write_bytecode = true` on `PySystemState` but this is potentially not ideal.

A possible fix for this would be to enhance the `haveCompiled` flag in `impl.loadFromSource` to check that the file is readable before attempting to open it, if it is not then can proceed to call `createFromSource`. Does this sound reasonable?

jeff5:
Thanks for diagnosing and reporting this,

The command line option `-B` will also suppress writing byte code.

Jython is mimicking CPython (2.x) behaviour in writing its compiled code adjacent to the source. The change that gives the files owner-only access responds to a security advisory about allowing users to replace files others then execute with their own rights. 

I'll admit I hadn't thought about the case where you share source. It would be better if the compiled code went to a per-user cache in the way we (now) have for indexes of JARs. (This is for the same reason.)

A look at the code suggests that if the shared directory were not writable, the write would fail harmlessly (with a log message at worst). There is clearly a performance penalty in re-compiling every time. If it were significant, perhaps you could allow a user privileged to write, to compile the files for everyone else (and fix the permissions).

Stewori:
Did you consider to compile the python sources to Java bytecode ahead of time?

```
java -cp whatever org.python.util.jython -c "import compileall; compileall.compile_dir('some_path')"
```
I am not sure how this would set the permissions of the created files, but you should be able to edit the permissions afterwards to your liking before deployment.

joeshannon:
[...]

PR #128, opened 2021-09-08
Support PYTHONDONTWRITEBYTECODE environment variable (GH-93)

joeshannon:
To enable a specific workaround for issue 93.


src/org/python/util/jython.java
```diff
@@ -723,6 +723,11 @@ private static void addDefaultsFromEnvironment(Properties registry) {
         // Go interactive after script. (PYTHONINSPECT because Python scripts may set it.)
         addDefault(registry, "python.inspect", getenv("PYTHONINSPECT"));
 
+        // PYTHONDONTWRITEBYTECODE
```
joeshannon, line 727: I put this in addDefaultsFromEnvironment as other env vars are read here but might be better elsewhere as it is not setting a registry entry.

jeff5: Let's do it here.

jeff5:
When I last re-worked `jython.java` I inferred that the architectural idea, before my time, had been that environment variables, like command line options, were only to be examined in `org.python.util.jython`, and that everything else should read the registry. This makes sense when you consider that `o.p.u.jython` is just one of the ways an interpreter can be started. I removed some use by `PySystemState` of environment variables, and made it go via the registry instead.

I'm not sure where `Options` was supposed to fit in. `o.p.u.jython` was setting options directly, but also `Options` was reading the registry.

I think a purist would unset `python.writebytecode` in response to `-B` or `PYTHONDONTWRITEBYTECODE` and `sys.dont_write_bytecode` would pick it up from there. However, doing that consistently means a lot of other change, so let's be pragmatic.

jeff5:
Can I bother you to sign a [Contributor Agreement](https://www.python.org/psf/contrib/contrib-form/)? We release under the PSF licence, but because our repos are not in the official Python GitHub organisation (because we were here first!) we cannot yet automate it with the same CLA robot. It asks for a bugs.python.org username, but hat has just been decommissioned and what is needed is your GitHub name @joeshannon .

joeshannon:
Thanks for looking at this. I have just signed the CA.

jeff5:
Thanks Joe. I'll look at 129 too.

PR #129, opened 2021-09-08
Only attempt to open source/bytecode if readable (GH-93)

joeshannon:
Previously if a compiled bytecode file existed but could not be read,
the import would fail (from the makeStream method) even if the source
file was present and it was readable.


joeshannon:
This allows to fallback to the source file if the compiled file cannot be read.

One consideration is that for the case that both files are not readable we will get `ImportError: No module named <...>` which is potentially confusing as it is not clear that it is a permissions issue.

jeff5:
@joeshannon  Thanks for providing this and #128.

PR #155, opened 2022-03-07
fix: safely compare strings using `.equals` instead of `==`

thetric:
Correctly compare Java Strings using `.equals()`/`switch case` instead of `==` (object identity)

### DIFF
--- a/src/org/python/core/PyBytecode.java
+++ b/src/org/python/core/PyBytecode.java
@@ -3,6 +3,7 @@ package org.python.core;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+
 import static org.python.core.Opcode.*;
 
 public class PyBytecode extends PyBaseCode implements Traverseproc {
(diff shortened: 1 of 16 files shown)


Stewori:
I think the `==` comparisons are actually okay because Jython is working with interned strings. Not sure how the performance impact is on this front; the comparisons are very frequently executed code. Even less sure how the trade-off is with moving to switch/case statements. That said, TBH I suspect the difference in performance would be hardly measurable at all, but it is tedious to test.

thetric:
> I think the == comparisons are actually okay because Jython is working with interned strings.

Nice to know. On a first glance this looked like a bug to me. Do you think that this PR is still worthwhile?

Stewori:
The style improvements look good to me, but OTOH fixing things that are not broken is unnecessary risk. I'm leaving this for @jeff5 to decide. Personally, I would accept the style improvements that are not semantic-changing (i.e. not the changed `==` comparisons) if it were a separate PR.

> looked like a bug to me

I guess with such a bug in place Jython would hardly pass a single test, probably not even startup :) . This is only about performance. OTOH I see the point that it requires to constantly educate devs and users that identifiers must be interned. Maybe the docs are not sufficiently alarming on this. Then, "OTOOH", this is fairly internal API.

jeff5:
@thetric  Somewhat like @Stewori, I think parts of this are good and others maybe misunderstand where one is allowed to rely on a string having been interned. (It is only in certain places, but attribute look-up is one.)

The `String` case statement might be a performance improvement if there are more than a few clauses, as it is based on hash.

thetric:
> The style improvements look good to me, but OTOH fixing things that are not broken is unnecessary risk. I'm leaving this for @jeff5 to decide. Personally, I would accept the style improvements that are not semantic-changing (i.e. not the changed == comparisons) if it were a separate PR.

Okay, I'm closing this PR and create a new one with only the stylistic changes. Thanks for your feedback :)

PR #156, opened 2022-03-07
Replace custom SHA224 digest with Java's

thetric:
As of Java 8 SHA-224 is part of the JDK: https://openjdk.java.net/jeps/130


jeff5:
This looks like a sensible change: it is always best to use something supplied by the platform.

One misgiving is that we are removing a public class here. IMO it ought never to have been public if all it is doing is this ... so, might someone be depending on it?

jeff5:
@thetric  Although a small change, could I bug you to sign the [Python Contributor Agreement](https://www.python.org/psf/contrib/contrib-form/)?

thetric:
@jeff5 Okay, I have signed the Python Contributor Agreement

jeff5:
@thetric  : thanks for that, now merged.

PR #157, opened 2022-03-07
build: upgrade to ASM 9.3

thetric:


jeff5:
We only use `asm`, `asm-commons` and `asm-util`.

Taking into account the need to check provenance of binary dependencies we bundle, it is easiest if a project member actually does the change. However, the prompt to upgrade things is welcome, and it is useful to know someone has run tests with the later version, so thanks.

PR #158, opened 2022-03-07
build: update to bouncycastle 1.70

thetric:
See https://www.bouncycastle.org/releasenotes.html


jeff5:
It looks like the `jdk18on-1.71` JARs are right. I'll do that from within the project, because of the provenance checking.

We don't seem to use whatever moved to the `bcutil` JAR. TBH, the BC JARs don't really work when we try to bundle them because we break the signature. The parts we use to support SSL work, but that's quite limited. Serious users have to supply a version on their class path, or use the slim JAR.

PR #179, opened 2022-05-27
GitHub Actions to run codespell

cclauss:
% `codespell --count jython/NEWS --ignore-words-list="statics,sur"`
```
./jython/NEWS:510: suport ==> support
./jython/NEWS:566: sematics ==> semantics
./jython/NEWS:592: appropiate ==> appropriate
./jython/NEWS:601: Synchonize ==> Synchronize
./jython/NEWS:758: occured ==> occurred
./jython/NEWS:902: istead ==> instead
./jython/NEWS:972: overriden ==> overridden
./jython/NEWS:1015: occured ==> occurred
./jython/NEWS:1027: overide ==> override
./jython/NEWS:1043: overriden ==> overridden
./jython/NEWS:1055: attibutes ==> attributes
./jython/NEWS:1210: doesnt ==> doesn't, does not
./jython/NEWS:1228: funcion ==> function
./jython/NEWS:1381: occured ==> occurred
./jython/NEWS:1479: overriden ==> overridden
./jython/NEWS:1500: readded ==> read
./jython/NEWS:1764: documention ==> documentation
./jython/NEWS:1784: overriden ==> overridden
./jython/NEWS:1799: useable ==> usable
./jython/NEWS:1826: refrences ==> references
```
https://github.com/cclauss/jython-2/actions

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -13,4 +13,5 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - run: pip install codespell
-    - run: codespell . --ignore-words=Misc/codespell/words.ignore --skip="*.decTest,*.g,*.ht,*.java,*.pem,*.py,*.uue,*.xml"
+    - run: codespell . --count --ignore-words=Misc/codespell/words.ignore
+                       --skip="*.decTest,*.g,*.ht,*.java,*.pem,*.py,*.uue,*.xml"
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -14,7 +14,7 @@ jobs:
     - uses: actions/checkout@v3
     - run: pip install codespell
     - run: codespell . --count --ignore-words=Misc/codespell/words.ignore
-                       --skip="lib-python/*,*.decTest,*.g,*.ht,*.java,*.pem,*.py,*.uue,*.xml"
+                       --skip="./extlibs/*,./lib-python/*,*.decTest,*.g,*.ht,*.java,*.pem,*.py,*.uue,*.xml"
 
   codespell-java:
 
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,5 +1,6 @@
-# https://pypi.org/project/codespell will fix common misspellings in text files. It’s designed primarily
-# for checking misspelled words in source code, but it can be used with other files as well.
+# https://pypi.org/project/codespell will fix common misspellings in text files.
+# It’s designed primarily for checking misspelled words in source code, but
+# it can be used with other files as well.
 name: codespell
 on: [pull_request, push]
 jobs:
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,9 @@
+name: codespell
+on: [pull_request, push]
+jobs:
+  codespell:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - run: pip install codespell
+      - run: codespell --count jython/NEWS --ignore-words-list="sur"  # --skip="*.css,*.js,*.lock,*.po"
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -13,4 +13,4 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - run: pip install codespell
-    - run: codespell NEWS README.md README.txt LICENSE --ignore-words=Misc/codespell/words.ignore
+    - run: codespell . --ignore-words=Misc/codespell/words.ignore --skip="*.g,*.ht,*.java,*.py,*.xml"
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,3 +1,5 @@
+# https://pypi.org/project/codespell will fix common misspellings in text files. It’s designed primarily
+# for checking misspelled words in source code, but it can be used with other files as well.
 name: codespell
 on: [pull_request, push]
 jobs:
@@ -6,4 +8,4 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - run: pip install codespell
-      - run: codespell --count ./NEWS --ignore-words-list="statics,sur"  # --skip="*.css,*.js,*.lock,*.po"
+      - run: codespell --count ./NEWS --ignore-words-list="statics,sur" --skip="*.java"
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -6,4 +6,4 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - run: pip install codespell
-      - run: codespell --count jython/NEWS --ignore-words-list="statics,sur"  # --skip="*.css,*.js,*.lock,*.po"
+      - run: codespell --count ./NEWS --ignore-words-list="statics,sur"  # --skip="*.css,*.js,*.lock,*.po"
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,12 +1,14 @@
-# https://pypi.org/project/codespell will fix common misspellings in text files.
-# It’s designed primarily for checking misspelled words in source code, but
-# it can be used with other files as well.
+# Check text for common misspellings and report them (GitHub Action)
+# https://pypi.org/project/codespell
+
 name: codespell
+
 on: [pull_request, push]
+
 jobs:
   codespell:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


NEWS
```diff
@@ -1497,7 +1497,7 @@ Jython 2.2 beta1
     - new-style classes and their instances are Java serializable
     - most new-style builtin types can be pickleds
     - __slots__ for new-style classes
-    - xml modules from pyxml readded
+    - xml modules from pyxml read
```
jeff5: I think the author meant "added back", but it was too long ago to make research worthwhile.

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,9 @@
+name: codespell
```
jeff5, line 4: I like to put a line or two of comment on an action, for those who won't know what `codespell` is doing.

jeff5 (changes requested):
Good, I think, with the tiny changes noted.

It could usefully be extended to other files. Locally I have tried `codespell` on some Java. I was disappointed it didn't know the difference between comment and code.

NEWS
cclauss:
```suggestion
    - xml modules from pyxml re-added
```

.github/workflows/codespell.yml
cclauss:
```suggestion
# https://pypi.org/project/codespell will fix common misspellings in text files. It’s designed primarily
# for checking misspelled words in source code, but it can be used with other files as well.
name: codespell
```

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,9 @@
...
+on: [pull_request, push]
+jobs:
+  codespell:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - run: pip install codespell
+      - run: codespell --count ./NEWS --ignore-words-list="statics,sur"  # --skip="*.css,*.js,*.lock,*.po"
```
cclauss:
```suggestion
      - run: codespell --count ./NEWS --ignore-words-list="statics,sur" --skip="*.java"
```

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,11 @@
+# https://pypi.org/project/codespell will fix common misspellings in text files. It’s designed primarily
+# for checking misspelled words in source code, but it can be used with other files as well.
```
cclauss:
```suggestion
# https://pypi.org/project/codespell will fix common misspellings in text files.
# It’s designed primarily for checking misspelled words in source code, but
# it can be used with other files as well.
```

jeff5 (approved):

jeff5:
I have a couple of tweaks, easier to make than explain, so I'll pull your PR branch and push a couple of sets. It would be nice to include source files (so as to spell-check the comments, especially the JavaDoc) but `codespell` produces too many false alarms.

I also want to test whether a spelling mistake is "red", which affects how many files it is sensible to include.

cclauss:
Makes sense...Small moves.  If you run with the `-w` flag it will write its guess but as we have already seen, that can be dangerous.

cclauss:
The `--skip="*.java"` flag is on so it should ignore Java files for now.  You can add others as well.

jeff5:
> The `--skip="*.java"` flag is on so it should ignore Java files for now. You can add others as well.

I believe that has no effect currently because the one file specified doesn't match the glob. I tried letting it find files by itself from `.`. You can `--skip` a list of folders and file-globs that lead to false alarms, but there are too many. It's a shame: amongst the false alarms are some genuine misspellings and this is a lot of work just for NEWS and a few other files.

> If you run with the `-w` flag it will write its guess.

Definitely one to keep for local use, and with `-i`.

.github/workflows/codespell.yml
```diff
@@ -7,8 +7,10 @@ on: [pull_request, push]
...
 jobs:
   codespell:
+
     runs-on: ubuntu-latest
+
     steps:
     - uses: actions/checkout@v3
     - run: pip install codespell
-    - run: codespell NEWS, README.md, README.txt, LICENSE --ignore-words=Misc/codespell/words.ignore
+    - run: codespell NEWS README.md README.txt LICENSE --ignore-words=Misc/codespell/words.ignore
```
cclauss: Maybe try `--skip="*.g,*.ht,*.java,*.py,*.xml"` to rapidly get to a reasonable subset of all files.

jeff5:
Ok, so I learned:

1. not to separate file names with commas on Unix, and that
2. a spelling mistake is a hard failure.

In view of the latter, it makes sense not to run this automatically on source code until `codespell` learns to tell the difference between variable names and English text. It would be annoying to have PRs look bad because something that isn't meant to be English matches in `codespell`.

I think `codespell` has a place in the development environment as a batch search mechanism, but the dictionary of troublesome names would take a lot of maintaining.

.github/workflows/codespell.yml
jeff5: But then we're filtering out most of the places this is could be useful to improve the readability of comments and maybe console messages.

cclauss:
`codespell || true` will make it a soft failure.

.github/workflows/codespell.yml
cclauss: It is better to have guardrails in place for those files that should only contain English text than to have no guardrails at all.

jeff5:
The sort of thing that might be useful for improving comments, documentation, etc. is:
```
codespell --count Lib --ignore-words=Misc/codespell/Lib.ignore
```
where `Misc/codespell/Lib.ignore` is currently:
```
ans
ba
cant
fo
hist
nd
nin
ois
spawnve
te
```
This gets the count down to 122, the majority of which look like genuine spelling mistakes, although some is echt deutsch. I think one could proceed in this way with selected sections of the code base, each with its own `.ignore` file. I'll add this.

It's potentially useful in the dev environment, not as a PR check because names in the code could become reasons for failure and then a contributor has to understand how to exempt their variable names to get a green merge button.

But for now, getting a 2.7.3b out and returning to Jython 3, beats fixing historic spelling mistakes. (I might start applying this in the new bits of Jython 3, though.)

.github/workflows/codespell.yml
jeff5: Yes, but it is easier to list them those to make globs for all the others.

cclauss:
OK.  Now we are down to 25 typos which should be more manageable to correct.

jeff5:
The original PR was to spell-check NEWS.

It is worth extending to named other files known to be (approximately) English text. But we should positively select the ones we want to check, not try to glob away everything else. (It is as likely someone will invent a new file extension as that they will introduce a new file of English.)

The files in `lib-python` and `extlibs` are not ours to edit, so really we're just down to my deliberate mistake now.

cclauss:
https://github.com/cclauss/jython-2/actions/runs/2408461306

cclauss:
https://github.com/cclauss/jython-2/actions/runs/2408471193

jeff5:
@cclauss : I thought this was good at the point it applied to specified text files. And the fixed typos in NEWS are welcome. Would you object if I reverted it to that point and pushed a change to remove the deliberate mistake?

I would not want to run the Java test (or a Python one) as an action: few will read the transcript of a passing test and we don't have time to investigate the alleged typos in all our source files. When the mood strikes, one can run ` codespell`  locally.

cclauss:
You can change this PR in any way that you see fit.

jeff5:
I rolled this back to something like the original PR, but applied to all the English text that seemed worthwhile. (Not `/Doc` because #180.) I'll do the more complicated things I had in mind on another change.

jeff5:
@cclauss : FYI, I made the more complicated idea that emerged from this discussion, using ignore-lists in `./Misc`, into https://github.com/jython/jython/commit/0f2f69894fba1ec9cf19ffa584b64e5590b2996f. Deferring actual typo fixes for a rainy day.

Thanks for the original PR.

cclauss:
`Misc/codespell/src.ignore`... Yes, that is what I was thinking for the separate jobs by language so we would have:
`Misc/codespell/html.ignore` for the `**/*.html` job
`Misc/codespell/java.ignore` for the `**/*.java` job
`Misc/codespell/python.ignore` for the `**/*.py` job

jeff5:
Splitting by different sub-tree worked better because what is a variable name in place one could be a typo in the other. The problem is not so much language as subject. `src` is mainly Java but there are some Python files.

At the moment though, I don't see spelling mistakes as a blocker for 2.7.3b

In the Gradle build (Jython 3) the tree is divided by both subject and language, so maybe:
`./core/.codespell/main.java.ignore`
`./core/.codespell/test.pythonExample.ignore`
`./build-tools/.codespell/python.ignore`

PR #201, opened 2022-07-26
Support vararg constructors

tpoliaw:
Checking if constructors are vararg methods allows autoboxing to be used for classes such as
```java
public class Foo {
    public Foo(String name, int... numbers) {
        // ...
    }
}
```

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -19,9 +19,9 @@ New Features
       variable (GH-128).
 
     - Under PR GH-101, the matching of sequence arguments to the parameters of
-      Java methods is improved. A Java varargs parameter does not match a final
-      sequence in the call if an array parameter could do so. (Thanks Peter
-      Holloway.)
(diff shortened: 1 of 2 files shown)


jeff5:
I was petty sure without reading it that we wouldn't merge this in v2.7.3 as it missed the beta and there isn't a b2.

It's so small and innocent-looking though, and the natural sequel to #101, that I'm tempted to risk it in the rc1. I'll take a look.

jeff5:
We'll ignore the codespell failure. (It learned some new misspellings while we were not looking.)

tpoliaw:
Thanks for looking at this, and accepting it after the beta release.

tpoliaw:
Always the most innocent-looking changes that catch you out. I shot myself in the foot with this one as it exposed #221.

Issue #207, opened 2022-08-28
java.lang.IllegalArgumentException: URI is not hierarchical


bts-webber:
I'm using Jython to develop Flink Java UDF. When I run it, I get an error:

> org.apache.flink.runtime.taskmanager.Task                    [] - Source: TableSourceScan(table=[[vvp, default, dwd_yunsec_host_aegis_proc_inner_s ... with job vertex id 717c7b8afebbfb7137f6f0f99beb2a94 (6/20)#0 (230a68259cc3877ac6b2be2ec4d7d299) switched from INITIALIZING to FAILED with failure cause: java.lang.IllegalArgumentException: URI is not hierarchical
> 	at java.io.File.<init>(File.java:420)
> 	at org.python.core.PrePy.getJarFileNameFromURL(PrePy.java:427)
> 	at org.python.core.PrePy._getJarFileName(PrePy.java:362)
> 	at org.python.core.PrePy.getJarFileName(PrePy.java:345)
> 	at org.python.core.PySystemState.doInitialize(PySystemState.java:1195)
> 	at org.python.core.PySystemState.initialize(PySystemState.java:1130)
> 	at org.python.core.PySystemState.initialize(PySystemState.java:1085)
> 	at org.python.core.PySystemState.initialize(PySystemState.java:1080)
> 	at org.python.core.PySystemState.initialize(PySystemState.java:1075)
> 	at org.python.core.PySystemState.initialize(PySystemState.java:1070)
> 	at org.python.core.ThreadStateMapping.getThreadState(ThreadStateMapping.java:56)
> 	at org.python.core.Py.getThreadState(Py.java:1741)
> 	at org.python.core.Py.getThreadState(Py.java:1737)
> 	at org.python.core.Py.getSystemState(Py.java:1757)
> 	at org.python.util.PythonInterpreter.<init>(PythonInterpreter.java:100)
> 	at org.python.util.PythonInterpreter.<init>(PythonInterpreter.java:93)
> 	at org.python.util.PythonInterpreter.<init>(PythonInterpreter.java:70)
> 	at sangduo.factory.JythonObjectFactory.createObject(JythonObjectFactory.java:31)
> 	at sangduo.udfx.CmdlineNormal.CmdlineNormalTest(CmdlineNormal.java:28)
> 	at sangduo.udfx.CmdlineNormal.open(CmdlineNormal.java:32)
> 	at StreamExecCalc$9384.open(Unknown Source)
[...]

PR #208, opened 2022-08-28
fix issues #207

bts-webber:


src/org/python/core/PrePy.java
```diff
@@ -417,6 +417,8 @@ public static String getJarFileNameFromURL(URL url) {
                         // path contains the target class in a JAR (named in group 1).
                         // Make a file URL from all the text up to the end of group 1.
                         fileURI = new URL("file:" + path.substring(0, m.end(1))).toURI();
+                        // When URL is Relative,convert to Absolute path
+                        fileURI = new URI("file:"+(new File(fileURI.toString().substring(5))).getCanonicalPath());
```
jeff5: Gosh, is it really that simple? Thanks for finding it.

This is going a long way only to come back and start again. We ought to build `fileURI` only once. Can you try something like:
```
File jarFile = new File(path.substring(0, m.end(1)));
fileURI = new URI("file:"+jarFile.getCanonicalPath())
```

Secondly, this bug is present in 2.7.2 and not in 2.7.1 (probably) because I refactored some code and the tests didn't catch my mistake. Is there something to be added here:
https://github.com/jython/jython/blob/63359da881edcf549909ee0492edcc91bb1acf6e/tests/java/org/python/core/PySystemStateTest.java#L114
 that would fail without your change, and pass with it. In other words, can we have a test that makes sure no-one undoes your good work? (I think that's the right place to exercise the fix.)

bts-webber:
2.7.1 works well in my flink udf.

and ,I added [test case for relative path](https://github.com/jython/jython/pull/208/commits/943c216f9013531e7e28029536d09202d162044c)

jeff5:
This arrived after the freeze before 2.7.3, so it got a bit lost. Sorry about that. I've tagged it for 2.7.4, which I'm pulling together now.

I'll pull it down and evaluate it. You may see some pushes. (Point of technique: it is better to [make a branch on your own repo](https://devguide.python.org/#quick-reference) for a change and PR from that, but leave it for now.)

jeff5:
It fails for me on Windows because `getJarFileNameFromURL` returns `null`. The underlying reason is that `new URI()` throws `java.net.URISyntaxException: Illegal character in opaque part at index 7: file:C:\some_dir\some.jar`. The offending character is the first "\\".

URI syntax: https://docs.oracle.com/javase/8/docs/api/java/net/URI.html

I notice the test you added is only in the non-Windows path of `testGetJarFileNameFromURL_jboss`.

The reason why the original code ran `new URL(...).toURI()` must I think be because URIs are [poorly standardised on Windows](https://en.wikipedia.org/wiki/File_URI_scheme), and this idiom seemed safe across platforms.

I'll exercise this on Linux too so I can understand what ought to happen.

jeff5:
Merged from master so we get latest fixes and, in particular, the CI.

jeff5:
This breaks Jython for Windows, so it's not acceptable as it stands. Also, are the expectations of the test (resolution of a vfs URL relative to the current working directory) actually correct? https://github.com/jython/jython/issues/207#issuecomment-1951376057

PR #243, opened 2023-05-29
Hold import lock only in getCodecState to avoid deadlock

tedyu:
Recently we encountered deadlock w.r.t. python import.
```
"http-nio-8080-exec-9":
  waiting for ownable synchronizer 0x00000007bf84aa00, (a java.util.concurrent.locks.ReentrantLock$NonfairSync),
  which is held by "http-nio-8080-exec-7"

"http-nio-8080-exec-7":
        at org.python.core.PySystemState.getCodecState(PySystemState.java:357)
        - waiting to lock <0x00000007bf8492b0> (a org.python.core.PySystemState)
        at org.python.core.codecs.getDefaultEncoding(codecs.java:38)
```
 `http-nio-8080-exec-7` was calling imp.createFromCode() which held `SystemState().getImportLock()`. Then `http-nio-8080-exec-7` tried to call `PySystemState.getCodecState`, waiting on `PySystemState`'s lock.
```
"http-nio-8080-exec-9":
        at sun.misc.Unsafe.park(Native Method)
        - parking to wait for  <0x00000007bf84aa00> (a java.util.concurrent.locks.ReentrantLock$NonfairSync)
        at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireQueued(AbstractQueuedSynchronizer.java:870)
        at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:1199)
        at java.util.concurrent.locks.ReentrantLock$NonfairSync.lock(ReentrantLock.java:209)
        at java.util.concurrent.locks.ReentrantLock.lock(ReentrantLock.java:285)
        at org.python.core.imp.load(imp.java:719)
        at org.python.core.PySystemState.getCodecState(PySystemState.java:360)
        - locked <0x00000007bf8492b0> (a org.python.core.PySystemState)
```
When `http-nio-8080-exec-9` called `imp.load` which needed to lock the import lock, deadlock formed.

This PR changes the locking in `PySystemState#getCodecState`. We only take importLock.
With this change, the deadlock shown in the stack trace cannot happen.

### DIFF
--- a/ACKNOWLEDGMENTS
+++ b/ACKNOWLEDGMENTS
@@ -193,6 +193,7 @@ Jython in ways large and small, in no particular order:
     Joe Shannon
     Dominik Broj
     shoaniki (GitHub name)
+    Zhihong (Ted) Yu
 
 Local Variables:
 mode: indented-text
--- a/src/org/python/core/PySystemState.java
+++ b/src/org/python/core/PySystemState.java
@@ -368,7 +368,7 @@ public class PySystemState extends PyObject
                 if (codecState == null) {
                     codecState = new codecs.CodecState();
                     // we have the importLock locked
-                    imp.load("encodings", false);
+                    imp.load("encodings");
                 }
             } catch (PyException exc) {
                 if (exc.type != Py.ImportError) {
--- a/src/org/python/core/PySystemState.java
+++ b/src/org/python/core/PySystemState.java
@@ -37,6 +37,7 @@ import java.util.Set;
 import java.util.StringTokenizer;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
@@ -193,7 +194,7 @@ public class PySystemState extends PyObject
 
     private int checkinterval = 100;
 
-    private codecs.CodecState codecState;
+    private volatile codecs.CodecState codecState;
 
     /** Whether bytecode should be written to disk on import. */
     public boolean dont_write_bytecode = false;
@@ -360,15 +361,21 @@ public class PySystemState extends PyObject
         return WinVersion.getWinVersion();
     }
 
-    public synchronized codecs.CodecState getCodecState() {
+    public codecs.CodecState getCodecState() {
         if (codecState == null) {
-            codecState = new codecs.CodecState();
+            importLock.lock();
             try {
-                imp.load("encodings");
+                if (codecState == null) {
+                    codecState = new codecs.CodecState();
+                    // we have the importLock locked
+                    imp.load("encodings", false);
+                }
             } catch (PyException exc) {
                 if (exc.type != Py.ImportError) {
                     throw exc;
                 }
+            } finally {
+                importLock.unlock();
             }
         }
         return codecState;
--- a/src/org/python/core/PySystemState.java
+++ b/src/org/python/core/PySystemState.java
@@ -37,7 +37,6 @@ import java.util.Set;
 import java.util.StringTokenizer;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
--- a/src/org/python/core/imp.java
+++ b/src/org/python/core/imp.java
@@ -1008,12 +1008,12 @@ public class imp {
      * Load the module by name. Upon loading the module it will be added to sys.modules.
      *
      * @param name the name of the module to load
-     * @param needToLock true if importLock should be locked
      * @return the loaded module
      */
-    public static PyObject load(String name, boolean needToLock) {
+    public static PyObject load(String name) {
         checkName(name);
         ReentrantLock importLock = Py.getSystemState().getImportLock();
+        boolean needToLock = !importLock.isLocked();
         if (needToLock) {
             importLock.lock();
         }
@@ -1024,15 +1024,6 @@ public class imp {
                 importLock.unlock();
             }
         }
-    }
-    /**
-     * Load the module by name. Upon loading the module it will be added to sys.modules.
-     *
-     * @param name the name of the module to load
-     * @return the loaded module
-     */
-    public static PyObject load(String name) {
-        return load(name, true);
     }
 
     /**
--- a/src/org/python/core/imp.java
+++ b/src/org/python/core/imp.java
@@ -1008,17 +1008,31 @@ public class imp {
      * Load the module by name. Upon loading the module it will be added to sys.modules.
      *
      * @param name the name of the module to load
+     * @param needToLock true if importLock should be locked
      * @return the loaded module
      */
-    public static PyObject load(String name) {
+    public static PyObject load(String name, boolean needToLock) {
         checkName(name);
         ReentrantLock importLock = Py.getSystemState().getImportLock();
-        importLock.lock();
+        if (needToLock) {
+            importLock.lock();
+        }
         try {
             return import_first(name, new StringBuilder());
         } finally {
-            importLock.unlock();
+            if (needToLock) {
+                importLock.unlock();
+            }
         }
+    }
+    /**
+     * Load the module by name. Upon loading the module it will be added to sys.modules.
+     *
+     * @param name the name of the module to load
+     * @return the loaded module
+     */
+    public static PyObject load(String name) {
+        return load(name, true);
     }
 
     /**
--- a/src/org/python/core/imp.java
+++ b/src/org/python/core/imp.java
@@ -1013,16 +1013,11 @@ public class imp {
     public static PyObject load(String name) {
         checkName(name);
         ReentrantLock importLock = Py.getSystemState().getImportLock();
-        boolean needToLock = !importLock.isLocked();
-        if (needToLock) {
-            importLock.lock();
-        }
+        importLock.lock();
         try {
             return import_first(name, new StringBuilder());
         } finally {
-            if (needToLock) {
-                importLock.unlock();
-            }
+            importLock.unlock();
         }
     }
 


tedyu:
@jeff5 
Can you take a look at the PR ?

Thanks

tedyu:
cc @tpoliaw

Stewori:
In #244 you point out that this PR and (I assume) the underlying issue is Jython 2.7.1-specific, at least as far as it was tested. There has been some work on Jython's synchronization logic since 2.7.1. Please check if the issue has been resolved in 2.7.3 or current master.

While maybe workable, the proposed fix appears symptomatic to me. It suggests existence of a race condition that should be fixed properly. Resolving a race condition can be seriously involved, which is one more reason to verify it has not been already fixed.

datavisortedyu:
The current patch is based on master branch.

I have looked at deadlock related fixes in newer releases.
I don't see any `closer` in the stack trace we captured.

The scenario described above seems to exist in master as well.

tpoliaw:
Do you have a simple example that can reproduce the deadlock? As @Stewori said, this feels very much like papering over the symptom, especially with the arbitrary 'try five times then fail' approach. I'd be interested in looking in to what is causing the underlying issue but it would help to have a starting point to work from.

datavisortedyu:
@tpoliaw 
The cause for deadlock is written in description of this PR.
I went through the code in master branch several times and I think the scenario we observed can still happen.

From my observation, the deadlock fixes w.r.t. `Closer` were not related to the scenario. I don't see any closer in the stack trace.
A coworker is testing with 2.7.3 release to see if similar deadlock appears.

The current approach retracts the lock on `PySystemState` when potential deadlock is detected.
I am open to suggestion on alternative approach to fixing this bug.

tedyu:
@tpoliaw 
I have simplified the PR and updated the description.
Can you take another look ?

The current approach should be easy to reason about.

Thanks

datavisortedyu:
@tpoliaw 
When you have time, can you take another look ?

tests based on 2.7.2 passed with this PR.

tpoliaw:
>The cause for deadlock is written in description of this PR.

Yes, but not a way to reliably reproduce it. It looks a bit better at first glance but without being able to test it, it's difficult to understand where the fault is coming from.

datavisortedyu:
[import-deadlock.txt](https://github.com/jython/jython/files/11679195/import-deadlock.txt)

@tpoliaw 
Please see the attached thread dump. I replaced method calls from my employer's repo with `...`
You can see the deadlock toward the end of the dump.

My PR solves the deadlock.

Stewori:
@tpoliaw The issue seems to be that `imp.load(String name)` acquires a lock and then calls during some module initialization `PySystemState.getCodecState`, which is synchronized. OTOH `PySystemState.getCodecState` loads the `encodings` module via `imp.load`. Two threads call `imp.load` and `getCodecState` somewhat simultaneously and end up waiting for each other's lock.

What is not obvious to me is why `getCodecState` does load the `encodings` module - nothing is done with it. That must be a conventional side-effect, i.e. existence of codec state shall imply that encodings has been loaded. Annoyingly, that convention is not documented. So, maybe it would be an enlightening experiment to just remove that `imp.load("encodings")` and see what happens. Anyway, loading the encodings module lazily might be an unnecessary risk. Perhaps it should be loaded during `PySystemState` initialization or at some other "safe" place.

Finally, I do not comprehend why `getCodecState` is synchronized. It cannot be for loading of encodings module because `imp.load` does its own locking. Is it to avoid creation of multiple codec states? Then maybe, just that part should be encapsulated in a synchronized block, excluding the call to `imp.load`.

My suggestion:
refine syncronize bhavior of `getCodecState` to just guard creation of codec state, excluding `imp.load`. I.e. put ``if (codecState == null) {codecState = new codecs.CodecState();}`` into a synchronized block and nothing more.

Do tests pass with that modification?
Can the OP report back if that minimal fix is sufficient to avoid the deadlock?

datavisortedyu:
@Stewori 
When I composed the last PR (changing lock order), I already noticed that `getCodecState` doesn't have to be synchronized. Other synchronized methods of PySystemState are static.

I have updated this PR by removing stateLock. We can do so because we hold importLock in `getCodecState`.
Please note: `codecState` is volatile.

See if the PR is easier to reason about.
Thanks

datavisortedyu:
From https://github.com/jython/jython/actions/runs/5211434156/jobs/9408497421
```
[exec] test test_urllib2net failed -- Traceback (most recent call last):
     [exec]   File "/home/runner/work/jython/jython/dist/Lib/test/test_urllib2net.py", line 203, in test_sites_no_connection_close
     [exec]     req = urllib2.urlopen(URL)
     [exec]   File "/home/runner/work/jython/jython/dist/Lib/urllib2.py", line 154, in urlopen
     [exec]     return opener.open(url, data, timeout)
     [exec]   File "/home/runner/work/jython/jython/dist/Lib/urllib2.py", line 435, in open
     [exec]     response = meth(req, response)
     [exec]   File "/home/runner/work/jython/jython/dist/Lib/urllib2.py", line 547, in http_response
     [exec]     response = self.parent.error(
     [exec]   File "/home/runner/work/jython/jython/dist/Lib/urllib2.py", line 467, in error
     [exec]     result = self._call_chain(*args)
     [exec]   File "/home/runner/work/jython/jython/dist/Lib/urllib2.py", line 407, in _call_chain
     [exec]     result = func(*args)
     [exec]   File "/home/runner/work/jython/jython/dist/Lib/urllib2.py", line 654, in http_error_302
     [exec]     return self.parent.open(new, timeout=req.timeout)
     [exec]   File "/home/runner/work/jython/jython/dist/Lib/urllib2.py", line 435, in open
     [exec]     response = meth(req, response)
     [exec]   File "/home/runner/work/jython/jython/dist/Lib/urllib2.py", line 547, in http_response
     [exec]     response = self.parent.error(
     [exec]   File "/home/runner/work/jython/jython/dist/Lib/urllib2.py", line 473, in error
     [exec]     return self._call_chain(*args)
     [exec]   File "/home/runner/work/jython/jython/dist/Lib/urllib2.py", line 407, in _call_chain
     [exec]     result = func(*args)
     [exec]   File "/home/runner/work/jython/jython/dist/Lib/urllib2.py", line [556](https://github.com/jython/jython/actions/runs/5211434156/jobs/9408497421#step:6:557), in http_error_default
     [exec]     raise HTTPError(req.get_full_url(), code, msg, hdrs, fp)
     [exec] HTTPError: HTTP Error 403: Forbidden
```
I don't think my change would result in 403 error.

jeff5:
@tedyu thanks for finding this and for offering a solution. Also, @Stewori and @tpoliaw thanks for reviewing this. What you're writing makes good sense.

The force-pushing is confusing the picture for me. A regular push ought to be fine, then we'll squash-merge them all and everyone will think we got it in one. ;)

Concurrency is hard. As Stefan explains, the deadlock is caused (almost certainly) by having two locks taken in variable order. I think we have to ensure they are taken in a defined order and released in the reverse order. Or maybe it is one lock (the import lock) to rule them both.

datavisortedyu:
`maybe it is one lock (the import lock) to rule them both.`

That's the approach I take now.
Obtaining the import lock would make sure that we only create one `CodecState` instance.
Since we don't take lock on `PySystemState` instance any more, the deadlock shown in the description is gone.

datavisortedyu:
From Lib/test/regrtest.py 
```
        # test_urllib2net        # unexpected output makes this a failure to regrtest.py
```
Looks like this test is known to be flaky.

jeff5:
> From Lib/test/regrtest.py
> 
> ```
>         # test_urllib2net        # unexpected output makes this a failure to regrtest.py
> ```
> 
> Looks like this test is known to be flaky.

The network tests are all somewhat flakey. Unfortunately, you do have to test the networking. The failure is repeatable. I'd like to know what fails, but for now let's assume it isn't this PR.

datavisortedyu:
I looked at test output for `test_urllib2net` a few times.
Apart from the call stack, I don't see other information.

I wonder if there is verbose logging that can be turned on for this test.

Thanks

Stewori:
> I think we have to ensure they are taken in a defined order and released in the reverse order. Or maybe it is one lock (the import lock) to rule them both.

In principle, yes. However, I suggest to reconsider the need for locking/synchronizing in `getCodecState` entirely. I do not really see what it accomplishes (@jeff5 , do you know?). Why not simply get rid of the lazy loading of `codecs` module and instead load it during startup? Then the codecState object would be initialized in the PySystemState constructor and getCodecState would become a trivial getter. It is probably required by every non-trivial Jython run anyway. The locking mechanism is already much too complicated. If there is an opportunity to simplify, it should be taken.

datavisortedyu:
If we actively load codecs module in `PySystemState` ctor, the ctor would incur the cost of taking import lock.
Possibly the original author of `getCodecState` intentionally avoided this startup cost.

With this PR, the loading of codecs module is more efficient than the previous formation (due to one fewer lock taken).

Cheers

jeff5:
> I looked at test output for `test_urllib2net` a few times. Apart from the call stack, I don't see other information.
> 
> I wonder if there is verbose logging that can be turned on for this test.
> 
> Thanks

One can run the regression tests in the dev environment locally (and should, before committing), with `ant regrtest` for the whole suite. Having built Jython once, one can run a single test with (for example):
```
dist\bin\jython -m test.urllib2net
```
or
```
dist\bin\jython -m test.regrtest -u network test_urllib2net
```
`regrtest` takes various options the better to control the test.

This fails for me on `master`, so something has broken while our backs were turned, and it's nothing to do with your PR.

jeff5:
> If we actively load codecs module in `PySystemState` ctor, the ctor would incur the cost of taking import lock. Possibly the original author of `getCodecState` intentionally avoided this startup cost.

I doubt that the time in taking a lock is significant and argues anyway for @Stewori's idea of just loading the codec registry once, since the lock is paid for on every `getCodecState`. I'd need to dig a bit to have an opinion whether it is true that: 

> It is probably required by every non-trivial Jython run anyway.

It's probably true on my machine, where an interest in internationalisation has led to a diverse combination of file system and console encodings. The registry is an instance field, so the cost of creating it is paid each time we create a `sys` module, but often this is only once.

@datavisortedyu : getting rid of `synchronized` is correct, but the polling and waiting, and the business with the boolean all seems unnecessary to me. Probably `volatile` is unnecessary too, but I always doubt my understanding of `volatile`. :) I think the problem is solved if `getCodecState` simply takes the corresponding import lock in the usual blocking way (and returns it at the end).

Our handling of concurrency in the core classes means well, but has no explicit policy behind it that I know of. I think there are problems in wait. For example, we can make getting a reference to the codec registry thread safe and properly synchronised. But the object itself looks unsafe for concurrent access. I'd rather get this right in 3 than fix it here.

datavisortedyu:
`but the polling and waiting`
I have pushed an update and simplified this part of code.

With the `volatile` variable, we only pay the cost of locking once in `getCodecState`. After the field is initialized, there is no more need to lock `importLock`.

jeff5:
I fixed (disabled) the regression test that fails in https://github.com/jython/jython/commit/8d4ad5f222c94c4ad674131c38d26d7eaf7d5cd1 , so if @datavisortedyu merges from upstream, tests should become more useful. Of course, they can always be run locally.

We'll squash merge when this finally makes it. Force pushing just clouds the issue.

In aid of testing:
> It is probably required by every non-trivial Jython run anyway.

I ran a few command lines under the debugger. (`..\remote.bat` runs Jython under the debugger for remote debugging from the IDE.) This calls `getCodecState`:
```
..\remote.bat -c "a=42"
```
Interestingly, it calls it with the import lock held at a depth of 6 times (re-entrantly). This doesn't call `getCodecState`:
```
..\remote.bat -S -c "a=42"
```
So basically the site module guarantees we call it, but other things will too of course.

I think the PR is still trying too hard to avoid taking a lock once we've decided we have to create the codec state and import `encodings`. I stepped into taking a lock a few times with the debugger. It is very few steps if the lock is not already held, and only a few more to check the current thread holds it. It is only slow when the current thread has to wait. I am against adding to the API of `imp` in pursuit of a time-saving that we have not shown is significant. 

In general [the double-checked lock pattern is broken](https://www.cs.umd.edu/%7Epugh/java/memoryModel/DoubleCheckedLocking.html). I believe `synchronized`, removed by the PR, is intended to serialise access to `codecState`, and `codecState` is now accessed outside of any lock. I believe the use of `volatile` to be a correct alternative. This blog from [CS6120 at cornell.edu](https://www.cs.cornell.edu/courses/cs6120/2019fa/blog/double-checked-locking) is helpful for pointing out that `volatile` isn't free, but I understand it to be mostly harmless for data that are mostly only read, which we have here. So I think this bit is good.

Footnote: other business takes me away from Jython for the next few weeks.

tedyu:
@jeff5 
I have simplified the change in `imp#load`. Turns out that we can use `isLocked` method to detect whether `imp#load` should lock importLock.
This way the `imp#load` API doesn't need to introduce additional parameter.

See if you have any more comment.

Thanks

BTW, when I did regular `git push` (commit e20796cb2aad25e3084849b50796cd1f11c017b0), the PR was auto-closed.
Probably this was the reason I used force push previously.

datavisortedyu:
@tpoliaw @Stewori 
I have addressed @jeff5 's review comment.
Can you take another look ?

thanks

Stewori:
I don't see why the change in `imp.load` would be required. It should be reverted to its original state. Since both threads would be accessing the same reentrent lock there should not be a possibility for deadlock. The second thread waits for whichever thread happens to get the lock first. That's it. The change in `PySystemState` looks okay to me.

datavisortedyu:
@Stewori 
I have taken out the change to `imp.java`

Please take another look.

datavisortedyu:
@tpoliaw 
Can you take a look ?

Stewori:
Given that tests pass, @jeff5 if you have no objections, and assuming that the OP has tested that the initial issue is cured by this PR I would say it can be merged.

jeff5:
I agree with others this is good (almost).
- I think TimeUnit imay now be an unused import.
- Could author please sign the PSF contributor agreement on python.org (or confirm they  already signed up).
This second request is a bit of a chore as we aren't integrated yet. (Requested for some time.)

datavisortedyu:
@jeff5 
I have signed the PSF contributor agreement.
The unneeded import has been dropped.

Please take another look.

Stewori:
@datavisortedyu please feel free to add yourself to the list of contributors.

datavisortedyu:
@jeff5 @tpoliaw 
Please let me know if there is anything I need to do.

datavisortedyu:
@jeff5 
Gentle ping.

jeff5:
Sorry, but other business really does take me away from Jython for the next few weeks. Thanks for doing this. I'm sure we'll have this in the next release.

datavisortedyu:
@jeff5 
Three weeks passed.
Is the next release getting closer ?

jeff5:
Three weeks closer. I haven't tried to name a date, but I think "autumn".

I'll pull this down to my desktop as I should try it there before just squash-merging. 

The approach here is satisfactory but I was left wondering what we should do if the import lock had not been accessible to take.

Suppose we were adding a feature that:
- needed to serialise calling threads with its own lock,
- might be called with the import lock held, and
- might import a module (e.g. as here on a cache miss).

It would not always be appropriate to borrow the import lock, so in general the feature must have its own, but then one cannot define the order the locks are taken. In Python 3, each module has its own import lock. I *think* this makes it less likely the problem will arise, but maybe harder to solve if it does.

datavisortedyu:
@jeff5 
Can this fix be merged into the next 2.7 release ?

Thanks

jeff5:
It is merged into `master`so it will be in the next release (Autumn: can't be more precise.).

datavisortedyu:
@jeff5 
August has come. I wonder if the next release would be done soon.
Thanks

Issue #211, opened 2022-09-28
httplib connections no longer support unicode


tpoliaw:
We have recently updated from 2.7.1 to 2.7.3 and have found a regression in how httplib handles unicode

```
import httplib

def post(data):
    try:
        conn = httplib.HTTPSConnection('httpbin.org')
        conn.request('POST', '/post', data)
        resp = conn.getresponse()
        print(resp.read())
    finally:
        conn.close()


post('this is a string')
post(u'this is unicode')
```

Running this with jython 2.7.1 gives the response for both calls. For 2.7.3, the first is successful but the second fails with 
```
Traceback (most recent call last):
  File "/tmp/http_demo.py", line 14, in <module>
    post(u'this is unicode')
  File "/tmp/http_demo.py", line 6, in post
    conn.request('POST', '/post', data)
  File "/tmp/jython-standalone-2.7.3.jar/Lib/httplib$py.class", line 1082, in request
  File "/tmp/jython-standalone-2.7.3.jar/Lib/httplib$py.class", line 1122, in _send_request
  File "/tmp/jython-standalone-2.7.3.jar/Lib/httplib$py.class", line 1078, in endheaders
  File "/tmp/jython-standalone-2.7.3.jar/Lib/httplib$py.class", line 898, in _send_output
  File "/tmp/jython-standalone-2.7.3.jar/Lib/httplib$py.class", line 870, in send
  File "/tmp/jython-standalone-2.7.3.jar/Lib/ssl$py.class", line 740, in sendall
  File "/tmp/jython-standalone-2.7.3.jar/Lib/_socket$py.class", line 1459, in meth
  File "/tmp/jython-standalone-2.7.3.jar/Lib/_socket$py.class", line 0, in sendall
TypeError: cannot make memory view because object does not have the buffer interface
```
This appears to be a consequence of #185 and PyUnicode [only superficially implementing `BufferProtocol`](https://github.com/jython/jython/blob/26731ea801a15feb86955e5284e61b8fb71ee306/src/org/python/core/PyUnicode.java#L183).

krismarc:
Any update on this? I have just experienced the same. If I convert the payload to bytes then it works "bytes(payload)".
However, currently I am not sure how to deal with.
```
url = httplib.HTTPConnection(host)
url.endheaders()
```

```
Traceback (most recent call last):
  File "/opt/wlp/integration/deployment/sr-wlp-deploy/utilities/sr_http_util.py", line 169, in http_invoke
    url.endheaders()
[...]

PR #256, opened 2023-06-17
Implicitly Convert Unicode to String for send and sendall

### LINKED ISSUE #254: Regression in 2.7.3 in `socket.socket.sendall` for sending Unicode
# Description

In Jython 2.7.3, passing a unicode string to `socket.socket.sendall` causes a TypeError to be raised saying a memory buffer could not be created. This seems to have been introduced in https://github.com/jython/jython/commit/e7413a64db501995c68229ec7981bd21a928a004 when `sendall` was made to not be an alias for `send`.

https://github.com/jython/jython/blob/8d4ad5f222c94c4ad674131c38d26d7eaf7d5cd1/Lib/_socket.py#L1218-L1223

The subtle reason why this fails is Jython doesn't implicitly convert unicode to string for `sendall()`, while CPython 2 does.

Tested that sending unicode to `sendall` does work in Jython 2.7.1. Sending unicode to `sendall` also works in CPython 2.7.

# Steps to Reproduce

Run the following code in Jython 2.7.3 and observe the error.
A second Python process must also be created to communicate with it.

```
Jython 2.7.3 (tags/v2.7.3:5f29801fe, Sep 10 2022, 18:52:49)
[OpenJDK 64-Bit Server VM (Eclipse Adoptium)] on java20.0.1
>>> import socket
>>> listen_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
>>> listen_socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, True)
>>> listen_socket.bind(('', 8888))
>>> listen_socket.listen(1)
>>> connection, address = listen_socket.accept()
>>> connection.sendall(u"hello\n")
  File "/path/to/jython-standalone-2.7.3.jar/Lib/_socket$py.class", line 1459, in meth
  File "/path/to/jython-standalone-2.7.3.jar/Lib/_socket$py.class", line 0, in sendall
TypeError: cannot make memory view because object does not have the buffer interface
>>> 
```

Running the same code in Jython 2.7.1 works as expected:

```
Jython 2.7.1 (default:0df7adb1b397, Jun 30 2017, 19:02:43) 
[Java HotSpot(TM) 64-Bit Server VM (Oracle Corporation)] on java1.8.0_181
>>> import socket
>>> listen_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
>>> listen_socket.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, True)
>>> listen_socket.bind(('', 8888))
>>> listen_socket.listen(1)
>>> connection, address = listen_socket.accept()
>>> connection.sendall(u"hello\n")
6
>>> 
```

Example code to run on another Python instance to connect to the port so the behavior can be seen.
Run the code above up to the `accept` line, then run this code in another process. Can be any version of Jython or Python for that matter.

```
Python 3.11.3 (main, Apr  7 2023, 19:29:16) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
>>> import sys, socket
>>> listen_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
>>> listen_socket.connect(('', 8888))
>>> try:
...   stdout = sys.stdout.buffer # Python 3
... except AttributeError:
...   stdout = sys.stdout # Python 2
...
>>> while 1:
...   stdout.write(listen_socket.recv(1))
...   stdout.flush()
... 
```

## Notes

This seems to be the same issue as #211, but just manifesting directly in the socket library.

In that issue, there was a discussion about if the Python specification requires `sendall` and related commands
to accept unicode, since the [docs](https://docs.python.org/2.7/library/socket.html) only every refer to the argument as a "string", but there is no encoding flag. Hence, I agree the _better_ approach programmatically is to set an encoding explicitly.

_However_, Python 2 silently upgrades strings to unicode, and considers `unicode` to be a subclass of `basestring`.  Therefore, I suggest library like `socket` should quietly convert unicode to the network interface because Python does consider them to be (base)"strings". I also would like to point out this _is_ the behavior of Jython prior to the breaking change.

TheMatt2:
Fixes #254 by converting unicode to string. This copies CPython in allowing unicode, even though the best practice is to explicitly only pass strings (or bytes) to socket.

Added test cases that are based on the existing cases for socket.


TheMatt2:
The test I added still sometimes fails with "port already in use".
However, this seems to also happen to the main socket test, which my test is closely based on.

ghost (approved):
my

jeff5:
Thorough test, thanks.

> The test I added still sometimes fails with "port already in use". However, this seems to also happen to the main socket test ...

Yes, my theory is that enthusiastic concurrency doesn't mix with the repeated opening and closing of the same port that occurs in the test. I have suppressed a few tests without quite stabilising it. It's not code I want to mess with that deeply.

jeff5:
@TheMatt2 : This looks good to me, thanks for offering it. I'd like to merge it.

I now have to bug you to **sign the PSF Contributor agreement** that we also use for Jython. If you have contributed to CPython, you'll have done this by an efficient process integrated into GitHub. No such luck here. We use this on-line form: https://www.python.org/psf/contrib/contrib-form/

Please confirm when you've done that.

I merged some changes from `master`, which include enabling (most of) `test_socket`,. It seemed like the right thing to do, in spite of some flakiness.

TheMatt2:
> Please confirm when you've done that.

I have signed the PSF agreement. Let me know if there is anything else I need to do on that front.

I spent some more time investigating the tests failing due to `error: [Errno 98] Address already in use`.
My main take away is it is possible for it to just work, it fails only sometimes. I may make this another PR if I figure out sometime useful.
Seems calling `gc.collect()` after closing the previous server connection helps, but that may be because it halts the program for over a second.

I think this PR can also close #211

Please merge when you think the time is right

jeff5:
Sorry for the lag -- distracted. I also think it probably closes #211, but I'll leave it open for now and hope to confirm that in the beta.

Thanks for this PR. I also think it probably closes #211, but I'll leave it open for now and hope to confirm that in the beta. I have added you to ACKNOWLEDGMENTS.

The flurry of pushes is for that and to ensure we're testing with the latest. (Still got the error 98, but second time was lucky.)  I'll squash-merge this shortly.

PR #261, opened 2023-07-09
CI: Set read-only permissions on GitHub Workflows

### LINKED ISSUE #241: CI: Set minimal permissions on GitHub Workflows
Hi!

I'm here to suggest that you set minimal permissions to your GitHub Workflows, because currently they don't specify the permissions for their jobs and their privileges are being determined by GitHub's defaults. Defining minimal permissions secures you against erroneous or malicious behaviours from external jobs you call from your workflow. It's specially important for the case they get compromised, for example.

It's a very simple change! I'd basically add 
```
permissions:
    contents: read
```

at the root of your workflows, and that would set a top-level read-only permission that would be inherited by any job that does not define job-level permissions.  It seems that currently none of your workflows require write permissions -- but in case you need them in the future, you'd keep the top-level read-only permissions and add the required write permissions as job-level.

Setting minimum permissions for workflows is recommended by [GitHub itself](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) and also by other security tools, such as [Scorecards](https://github.com/ossf/scorecard) and [StepSecurity](https://github.com/step-security).

Let me know what you think about this. If you agree, I'd be happy to contribute with a PR =).

#### Context
I'm Diogo and I work on Google's Open Source Security Team([GOSST](https://github.com/diogoteles08#about-gosst-ghost)) in cooperation with the Open Source Security Foundation ([OpenSSF](https://openssf.org/)). My core job is to suggest and implement security changes on widely used open source projects 😊

meenbeese:
Resolves #241 by defining a read-only permission for all workflow files at a root level.

### DIFF
--- a/.github/workflows/ant-javatest.yml
+++ b/.github/workflows/ant-javatest.yml
@@ -8,6 +8,8 @@ on:
   pull_request:
     branches: [ master ]
 
+permissions:
+  contents: read
 
 jobs:
 
--- a/.github/workflows/ant-regrtest.yml
+++ b/.github/workflows/ant-regrtest.yml
@@ -8,6 +8,8 @@ on:
   pull_request:
     branches: [ master ]
 
+permissions:
+  contents: read
 
 jobs:
 
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -5,6 +5,9 @@ name: codespell
 
 on: [pull_request, push]
 
+permissions:
+  contents: read
+
 jobs:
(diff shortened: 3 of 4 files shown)


jeff5:
Thanks. This implements the standard good practice we were recommended in #241 and hadn't got round to yet (and might have forgotten). As such, I won't bother you to sign the Python contributor agreement, but if you want to make further contributions that would come up.

As you've seen we don't run the CI on PRs without inspecting them first.

PR #264, opened 2023-07-25
Docs: Create Security Policy

### LINKED ISSUE #263: Create a Security Policy
Hi! I'm Diogo and I'm back (see #241) hoping to offer a bit more help with security enhancements.

This time I'm here to suggest that you expose a way that users can report eventual vulnerabilities in a safe and efficient way. This is usually done through a Security Policy, which is a GitHub standard document (SECURITY.md) added on the root of the repo and that will be visible to the users in the "Security Tab", as you can see bellow:

image

Having a Security Policy is a recommendation from [Github](https://docs.github.com/en/code-security/getting-started/securing-your-repository#setting-a-security-policy) itself, and from [Scorecard](https://github.com/ossf/scorecard/blob/main/docs/checks.md#security-policy) (being a security measure of medium priority).

Aiming to make this change easier, I'll take the liberty and submit a suggestion of a Security Policy as a PR. Please feel free to edit it directly or ask me for editions until it is in compliance with how jython would best handle vulnerability reports.

diogoteles08:
Closes #263

I've created the SECURITY.md file following a GitHub's template and considering that you'd request that users report vulnerabilities through the [security advisory](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing/privately-reporting-a-security-vulnerability), which is a handy new GitHub feature, but it's still in beta and **has to be manually enabled by a maintainer**.

If you're interested in this feature, you can activate it following this steps:

1. Click on this [link](https://github.com/jython/jython/settings/security_analysis) to go to Code security & analysis section on your repo's settings
2. Click "Enable" for "Private vulnerability reporting (Beta)"

In case you'd rather not use this feature, you can also request users to report vulnerabilities to an email. If that's the case, let me know which email you would like to receive the reports and I can submit the change.

Additionally, feel free to edit or suggest any changes to this document, it is supposed to reflect the amount of effort the team can offer to handle vulnerabilities.

Also, at the PR #261 it was mentioned that one might need to sign the Python CLA in order to contribute. If needed, can you guide me on how to do that? I haven't found the instructions.


jeff5:
Thanks for suggesting this. The PSF has its own mechanism, but I'm fairly sure what would happen these days if it were used for a v2.7, so this is a better route at least until we have a 3.x.

Yes, we should have you sign the PSF Contributor agreement that we also use for Jython. If you have contributed to CPython, you'll have done this by an efficient process integrated into GitHub. We use this on-line form: https://www.python.org/psf/contrib/contrib-form/

Please confirm here when you've done that.

Note to self:
- [x] Enable reporting in the repo settings.

SECURITY.md
```diff
@@ -0,0 +1,13 @@
+# Security Policy
+
+## Supported Versions
+
+Security updates are applied only to the latest release.
```
jeff5: ... or is it better to say "to the development tip, to appear in the next release"?

diogoteles08:
Signed it! Thanks for the instructions

SECURITY.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Reporting a Vulnerability
+
+If you have discovered a security vulnerability in this project, please report it privately. **Do not disclose it as a public issue.** This gives us time to work with you to fix the issue before public exposure, reducing the chance that the exploit will be used before a patch is released.
+
+Please disclose it at our [security advisory](https://github.com/jython/jython/security/advisories/new).
+
+This project is maintained by a team of volunteers on a reasonable-effort basis. As such, vulnerabilities will be disclosed in a best effort base.
```
jeff5: I didn't understand "vulnerabilities will be disclosed in a best effort base".

They'll be *handled* on a best-efforts basis ("addressed" maybe). Am I missing something different you intended to express?

*(After some research ...)* Ok, I see the process includes passing the information downstream. Mostly alerts to us are about a vulnerability in a dependency, so they have a CVE number and the solution is to update the dependency. I guess those need not propagate downstream?

In a few cases, what someone thought was a vulnerability, didn't appear to us as such (e.g. you can exploit X to make Y happen only if you have the privilege to do Y anyway).

We've few issues of our own making. CVE-2019-17664 is really us I think.

jeff5 (changes requested):
Could you help us be clearer about the process?

SECURITY.md
diogoteles08: What do you mean by "development tip"? Is it like the dev branch?

SECURITY.md
diogoteles08: > I didn't understand "vulnerabilities will be disclosed in a best effort base".
> 
> They'll be _handled_ on a best-efforts basis ("addressed" maybe). Am I missing something different you intended to express?
> 
> _(After some research ...)_ Ok, I see the process includes passing the information downstream.

Yep, the Security Policy has the purpose to let you know of possible vulnerabilities in your code so that you can properly evaluate it, plan on how to handle, and disclosure it as soon as possible. This disclosure is important to make your users aware of it and evaluate if they need/can avoid it. That said, I also think changing the text to `vulnerabilities will be disclosed and handled in a best effort base` would make sense. I can make this change if you wish =)

But note that until now I'm talking only about vulnerabilities specifically in your project, not in your dependencies.

> Mostly alerts to us are about a vulnerability in a dependency, so they have a CVE number and the solution is to update the dependency. I guess those need not propagate downstream?

In the case of a vulnerability discovered in your dependency, it would depend on how you pin that dependency. If you loosely pin it (e.g., `>=4.1.12`), then the users would already be safe and nothing needs to be done by you. 

However, if you strictly pin that dependencies (which *I think* it's the case for Jython), you'd need to ship a new version of the project with the updated dependency, so that your project doesn't require a vulnerable version. But even in this case, this shouldn't need to be disclosed nor propagated downstream; you could just let your users know on your release notes.

> In a few cases, what someone thought was a vulnerability, didn't appear to us as such (e.g. you can exploit X to make Y happen only if you have the privilege to do Y anyway).

That makes sense! It's a very possible use case. One of the purpose of the Security Policy is enable this private conversation between the reporter and the maintainers, so that you can also conclude that the vulnerability is not a real concern.

SECURITY.md
jeff5: I mean "the latest state of the `master` branch".

As it didn't mean that to you, it's the wrong phrase to use.

SECURITY.md
jeff5: Thanks, that's a good clarification. As a newcomer to the process (and maybe others will be) explicit text is better than technical shorthand. I didn't find the GitHub help that clear for the busy reader.

Good to know dependencies don't have to involve us.

FYI, we take two approaches to dependencies:
1. Some appear shadowed in the JAR and others bundled in the installation kit. In that case, we have obviously chosen specific versions, and the advantage is we've tested with them.
2. We supply a JAR (the "slim" JAR) that contains only this project's work and we name all dependencies in the Gradle build which also generates a POM. The versions are explicitly those we tested with.

I much prefer approach 2, but I'm not sure how to pin versions loosely without undermining the validity of tests. I'm not a Maven ... maven. I wonder if some form of regular re-test with latest matching versions of everything would be a good precaution?

jeff5: I'll make the tweak to wording. I'm sure the policy file will evolve. What you've done is show us how to get started.

I'll add you to ACKNOWLEDGEMENTS and push that onto your PR branch.

jeff5 (approved):

jeff5:
@diogoteles08 : I'll give this a few days so you can find _my_ mistakes, but intend to squash-merge after that.

diogoteles08:
Hey @jeff5 , looks good to me! Good job and way to go

jeff5:
@diogoteles08 : thanks for showing us how this works. I've enabled it in the repo settings.

PR #267, opened 2023-08-16
Changing grammar for decorators that don't have parenthesis

tatianacv:
Processing as a `CALL`  instead of an `Attribute` for the case when decorators don't have parenthesis.

```python
@decomaker(argA, argB, ...)
def func(arg1, arg2, ...):
    pass
```
This is equivalent to:
```python
func = decomaker(argA, argB, ...)(func)
```
Ref: [PEP 318 – Decorators for Functions and Methods](https://peps.python.org/pep-0318/)

Therefore, using `@decorator` is like a `CALL`.

### DIFF
--- a/grammar/Python.g
+++ b/grammar/Python.g
@@ -482,7 +482,7 @@ decorator
       RPAREN
     |
       {
-          $etype = $dotted_attr.etype;
+          $etype = actions.makeCall($LPAREN, $dotted_attr.etype);
       }
     ) NEWLINE
     ;


jeff5:
Thanks for this @tatianacv.

I think you are probably correct about this nit in the grammar of **Jython 2** but you're proposing a merge into the very immature Jython 3 branch, not the default `master` branch where releasable versions grow. (We intend `main` be the default when Jython 3 reaches a useable state.)

It would be great if you were interested in that version, but  `Python.g` is dead code on that branch (pending final judgement), so I guess not.

I write "probably correct" because:

1. Generating a compiler is scary and I'd really like @fwierzbicki's opinion if it can be got.
2. There's no shortage of decorator use in the test suite, so how come this hasn't been a problem in the last 15 years?

Could you target `master` please, and could you give an example of something that doesn't work and is fixed by this change?

jeff5:
PS Good practice is to make a branch on your fork for a proposed change, then PR from that branch to (in this case) our `master`. 

The reason is that, when we reject it, or merge it by squashing or rebasing (which generally creates a new change set), your repo can be synchronised easily with ours with a pull. Also, it is possible this way to have multiple independent changes on the go.

tatianacv:
Thanks for the response. For point _"There's no shortage of decorator use in the test suite, so how come this hasn't been a problem in the last 15 years?"_, I think the reason why decorators with dotted attributes without parameters were not used frequently. Looking at the tests suite in `jython,` I don't see a test that includes that case. We were working with this example, and when we made the change in this PR, we processed it correctly.

```python
import tensorflow as tf

@tf.function
def func(a):
  return a

a = tf.range(5)
b = func(a)
```

tatianacv:
To not bring over all the commits made in `main` branch, I had to create a new PR that targets `master` branch. Please refer to https://github.com/jython/jython/pull/274

jeff5:
Thanks @tatianacv . Closing this one to avoid confusion.

PR #274, opened 2023-08-25
Changing grammar for decorators that don't have parenthesis

tatianacv:
Processing as a `CALL`  instead of an `Attribute` for the case when decorators don't have parenthesis.

```python
@decomaker(argA, argB, ...)
def func(arg1, arg2, ...):
    pass
```
This is equivalent to:
```python
func = decomaker(argA, argB, ...)(func)
```
Ref: [PEP 318 – Decorators for Functions and Methods](https://peps.python.org/pep-0318/)

Therefore, using `@decorator` is like a `CALL`.

Same PR as #267 but targeting `master` branch.

### DIFF
--- a/grammar/Python.g
+++ b/grammar/Python.g
@@ -488,7 +488,7 @@ decorator
       RPAREN
     |
       {
-          $etype = $dotted_attr.etype;
+          $etype = actions.makeCall($LPAREN, $dotted_attr.etype);
       }
     ) NEWLINE
     ;


jeff5:
Did you run any tests in your environment? You should at least run `ant regrtest` before commit.

I think you're onto something with the grammar, but it isn't ready yet as the regression tests fail.

There are two very similar grammars. One used for the file-compiler and one for the interactive prompt. For reasons I don't understand, they couldn't be a single grammar. So when you find a thing that works in `Python.g` please add it to ` PythonPartial.g`.

_Edited: The warning about AT isn't in fact new, so not a clue to the test failures._

jeff5:
It appears from these runs:
https://github.com/jython/jython/actions/runs/5977032306/job/16321066933?pr=274
https://github.com/jython/jython/actions/runs/5977032315/job/16321067192?pr=274
https://github.com/jython/jython/actions/runs/5977032315/job/16321067469?pr=274
that a simple decorator with no parentheses like `@classmethod` is not called with the correct argument. (I wonder why it worked for you with a dotted decorator `@tf.function`?) I think the call in question happens at definition time.

Just launching the interactive REPL ought to be enough to cause that. It seems to happen when `site.py` is imported, which it is by default. Or you could launch with `-S` and play directly with a decorator. You'll be using the grammar in `PythonPartial.g` when in the REPL.

Hopefully, this is a matter of calling the action correctly in the grammar, not delving into code generation, although looking at what is generated might be helpful.

grammar/Python.g
```diff
@@ -488,7 +488,7 @@ decorator
       RPAREN
     |
       {
-          $etype = $dotted_attr.etype;
+          $etype = actions.makeCall($LPAREN, $dotted_attr.etype);
```
jeff5, line 491: Use of `LPAREN` seems dubious here as we have not encountered a match of that kind.

Since `dotted_attr` is able to match either a simple name or an attribute expression (and returns a different AST node type in each case) I am beginning to think the problem might be in the processing of the decorators subsequently, that it overlooks the possibility of an attribute expression.

jeff5:
No, the problem is not with the grammar at all. Jython produces the correct AST (see below). The decorators go into an attribute of the function definition called `decorator_list` and we do this identically to CPython.

If it doesn't _run_ correctly, the problem is further downstream, maybe here:
https://github.com/jython/jython/blob/e55e057637eb9fa30710f4c6720186a37c2b8974/src/org/python/compiler/CodeCompiler.java#L538

Could you raise this as an issue instead of a PR, including any useful text from here and #267? Then it can be cited in any PR that claims to fix it. (Yours if you still have the appetite.)

It's always better to start with an issue for anything remotely complicated. That way competing analyses of the problem can all be in one place, and stay around until it is fixed, rather than getting submerged with each failed PR. And even if _you_ don't fix it, you contributed by starting the hunt. I should have sent you that way after the first failed PR.

-----
Here is a little program you can run in Python 2.7 or Jython. We get the same AST:
```python
# ast_deco.py: How the AST represents a decorated function definition
# Jython Issue/PR GH-274 (previously 267, but should be an issue).

import ast

# Key to names: m, d, a = module, decorator, argument
d="""\
@deco
def func(): pass
"""

md="""\
@mod.deco
def func(): pass
"""

da="""\
@deco(arg)
def func(): pass
"""

mda="""\
@mod.deco(arg)
def func(): pass
"""

for prog in (d, md, da, mda):
    p = ast.parse(prog, mode='exec')
    print
    print prog, "decorator_list = ",
    #print(ast.dump(p))
    print ast.dump(p.body[0].decorator_list[0])
```

grammar/Python.g
khatchad: > Use of `LPAREN` seems dubious here as we have not encountered a match of that kind.

@tatianacv can correct me if I am mistaken, but I think she just needed to send *something* in as an argument there.

khatchad: > Since `dotted_attr` is able to match either a simple name or an attribute expression (and returns a different AST node type in each case) I am beginning to think the problem might be in the processing of the decorators subsequently, that it overlooks the possibility of an attribute expression.

I believe that @tatianacv is modeling this change off of the [case where there is parentheses](https://github.com/jython/jython/pull/274/files#diff-ff477a8df29d93b68543d917d4e9b0e31d828d9e218cae41e10774443a20d5e2L480-L481). The idea, if I am not mistaken, is that, with decorators (as opposed to "normal" function calls), the following is semantically equivalent:

```python
@tf.function
def func():
    pass

@tf.function()
def func2():
    pass
```

Thus, they should be processed the same. That is what we are seeing in our client code, and that is the fixed being proposed. I agree that the argument is dubious, but I believe that there is no `makeCall()` that takes a single parameter. Perhaps an overloaded `makeCall()` should be part of this PR.

khatchad:
> If it doesn't _run_ correctly, the problem is further downstream

If the paren and non-paren cases are indeed semantically equivalent, then wouldn't the processing in the grammar files also be similar?

jeff5:
> If the paren and non-paren cases are indeed semantically equivalent, then wouldn't the processing in the grammar files also be similar?

I believe you are correct @khatchad  that it means the same thing in the end but the AST captures the syntactical difference. We try to do the same thing as CPython wherever the results are visible from Python. In this case, the output from the program above in CPython 2.7.18 and Jython 2.7.3 is identical:
```
PS 274> python ast_deco.py

@deco
def func(): pass
decorator_list =  Name(id='deco', ctx=Load())

@mod.deco
def func(): pass
decorator_list =  Attribute(value=Name(id='mod', ctx=Load()), attr='deco', ctx=Load())

@deco(arg)
def func(): pass
decorator_list =  Call(func=Name(id='deco', ctx=Load()), args=[Name(id='arg', ctx=Load())], keywords=[], starargs=None, kwargs=None)

@mod.deco(arg)
def func(): pass
decorator_list =  Call(func=Attribute(value=Name(id='mod', ctx=Load()), attr='deco', ctx=Load()), args=[Name(id='arg', ctx=Load())], keywords=[], starargs=None, kwargs=None)
```

The output is quite similar with Python 3.11, once you fix the `print` statements.

It looked like @tatianacv had the right answer to begin with. On closer inspection, the AST is correct already, so the problem is in what we do with the AST later.

Add that the regression tests all fail. (You can run these in your dev environment with `ant regrtest`.)

I'm surprised this didn't come up in your use of a patched version in your own work (which BTW is exciting to see).

grammar/Python.g
jeff5: > > Use of `LPAREN` seems dubious here as we have not encountered a match of that kind.
> 
> @tatianacv can correct me if I am mistaken, but I think she just needed to send _something_ in as an argument there.

It looks like `$LPAREN` would be `null` in practice. That may be tolerated in later processing (perhaps not used unless there is a syntax error in need of a location) but is ... dubious.

khatchad: @tatianacv Could you confirm that there is no `makeCall()` that takes only a single parameter that would match the argument `$dotted_attr.etype`?

khatchad:
(Edited to correct the code inclusion. -- Jeff)

> > If the paren and non-paren cases are indeed semantically equivalent, then wouldn't the processing in the grammar files also be similar?
> 
> I believe you are correct @khatchad that it means the same thing in the end but the AST captures the syntactical difference.

I see what you mean @jeff5. But, conceptually, `@tf.function`, ilke `@tf.function`, (for example) *is* a *call*. Thus, in *both* cases, it would (at least) *conceptually* make sense that both productions include `actions.makeCall()` --- because they are both calls. In other words, although (for example) `@tf.function` looks like an attribute reference, it's not. It's really a function call, like you would see in Ruby. I believe that this is a quirk with the Python grammar; AFAIK this is the only time you can "call" a no-arg function in Python without using empty parentheses (`()`).

> We try to do the same thing as CPython wherever the results are visible from Python. In this case, the output from the program above in CPython 2.7.18 and Jython 2.7.3 is identical:
> 
> ```
> PS 274> python ast_deco.py
> 
> @deco
> def func(): pass
> decorator_list =  Name(id='deco', ctx=Load())
> 
> @mod.deco
> def func(): pass
> decorator_list =  Attribute(value=Name(id='mod', ctx=Load()), attr='deco', ctx=Load())
> 
> @deco(arg)
> def func(): pass
> decorator_list =  Call(func=Name(id='deco', ctx=Load()), args=[Name(id='arg', ctx=Load())], keywords=[], starargs=None, kwargs=None)
>
> @mod.deco(arg)
> def func(): pass
> decorator_list =  Call(func=Attribute(value=Name(id='mod', ctx=Load()), attr='deco', ctx=Load()), args=[Name(id='arg', ctx=Load())], keywords=[], starargs=None, kwargs=None)
> ```

Again, I see what you mean @jeff5, but I think that the first two above are incorrect. Both `@deco` and `@mod.deco` above should be calls despite not having the parentheses. I would seem that CPython then has the same problem as Jython.

> The output is quite similar with Python 3.11, once you fix the `print` statements.
> 
> It looked like @tatianacv had the right answer to begin with. On closer inspection, the AST is correct already, so the problem is in what we do with the AST later.
> 
> Add that the regression tests all fail. (You can run these in your dev environment with `ant regrtest`.)

@tatianacv Can you investigate why the regression tests are failing?

> I'm surprised this didn't come up in your use of a patched version in your own work

Unfortunately, we are not running the Jython regression tests in our CI. We should probably add those.

>  (which BTW is exciting to see).

Thanks! We will (hopefully) be releasing it soon :).

jeff5:
> Unfortunately, we are not running the Jython regression tests in our CI. We should probably add those.

Not necessarily. They take ages and would only be useful if you have modified Jython.

However, if you _have_ modified Jython, and are thinking of submitting a PR then yes, you really ought to run them to check your work before doing so. But you would do that manually in your development environment.

When you're confident to submit a PR, the tests will run on a selection of platforms on the GitHub CI as you can see in this PR. A project member has to authorise the run because of the potential for abuse, but apart from that, it's automatic.

In this case, If I download your PR branch to my PC I can build it ok:
```
PS Jython-2> ant clean
...
PS Jython-2> ant
...
```
You don't need `clean` every time but it is a good idea after switching branches, or if you suspect the incremental build has poisoned the well. I could test it with `ant javatest` and  `ant regrtest` at this point but let's just see if it runs at all.
```
PS Jython-2> dist\bin\jython
Exception in thread "main" Traceback (most recent call last):
  File "...\Jython-2\dist\Lib\site.py", line 68, in <module>
    import os
  File "...\Jython-2\dist\Lib\os.py", line 434, in <module>
    import UserDict
  File "...\dist\Lib\UserDict.py", line 3, in <module>
    class UserDict:
  File "...\dist\Lib\UserDict.py", line 72, in UserDict
    @classmethod
TypeError: classmethod expected 1 argument, got 0
```
It's clearly not working because it can't execute the built-in `classmethod` decorator. This is happening in` site.py`, which is run automatically on startup. 

We can suppress running` site.py` with `-S` and get a prompt. But if I define and use the simplest possible decorator, I get the same error:
```
PS Jython-2> dist\bin\jython -S
Jython 2.7.4a1-DEV (heads/pr-274:ec0b1f20d, Sep 6 2023, 18:46:56)
[Java HotSpot(TM) 64-Bit Server VM (Oracle Corporation)] on java1.8.0_321
>>> def deco(f): return f
...
>>> @deco
... def f(): pass
...
TypeError: deco() takes exactly 1 argument (0 given)
>>> raise SystemExit
PS Jython-2>
```
Without `site.py`, `exit()` is undefined, so you have to `raise SystemExit` to exit.

jeff5:
Let's back up a bit.

@tatianacv: when you wrote in https://github.com/jython/jython/pull/267#issuecomment-1693435451, "when we made the change in this PR, we processed it correctly" I assumed you to mean the code didn't run correctly before the change in the PR, and with the PR change it did.

Dotted decorators work correctly for me.

- Could you show me "processed incorrectly" and why it is incorrect?
- Could you show me Jython doing the wrong thing when executing source code like that in the message?

jeff5:
Thanks for all this @tatianacv, @khatchad. My grasp on decorators is much stronger than when we started! Let me strengthen yours.

> But, conceptually, `@tf.function`, ilke `@tf.function`, (for example) _is_ a _call_.

Agreed the `'@' expression NEWLINE` construct compiles to a call. It will call `expression` supplying the result of the function defined immediately after as the single argument. We generate code for that call because we are processing a decorator application.

> Thus, in _both_ cases, it would (at least) _conceptually_ make sense that both productions include `actions.makeCall()` --- because they are both calls.

Ah, but
```
@decomaker(arg)
```
compiles to two calls. `decomaker(arg)` is an expression that happens to be call. So we will record a call in the AST with `actions.makeCall()` and subsequently generate code for that call, which is to say code to evaluate the expression. But evaluating that expression must _return_ the decorator that we subsequently call with the function defined immediately after as the single argument.

It was staring out of PEP 318 all along. It tells us
```
@decomaker(argA, argB, ...)
def func(arg1, arg2, ...):
    pass
```
is equivalent to:
```
func = decomaker(argA, argB, ...)(func)
```
Two calls, not one. The first needs `actions.makeCall()` to build it. The second is implied by the `@`.

The grammar fragment at issue is the grammar of just the `expression`, not its subsequent application to the function implied by the `@`. _Any_ expression will do, syntactically, but it will fail at run-time if it turns out not to be callable with a single argument.

I think there is a fault in our grammar that it does not accept an expression in general, which the reference implementation would. I don't expect us to fix that in 2. (If you need it, raise an issue.) Currently we expect to use the Python 3 PEG grammar verbatim in Jython 3, so it would automatically be the same.

jeff5:
I think I've recognised that you're not actually using Jython, just our grammar in some way. You thought it should be different, because of a misunderstanding, hopefully corrected. But that change fails all the tests. I'll close this PR as invalid.

Please feel free to continue to use our work, and good luck with the research.

PR #275, opened 2023-08-26
Updgraded guava to guava-32.1.2-jre.jar to address CVE-2023-2976

bhasuanant:
guava-32.1.2-jre.jar addresses CVE-2023-2976


Stewori:
Given that tests pass and since this is a more or less "trivial" PR I think we can accept it without requiring signed contributor agreement (okay for you, @jeff5 ?). I verified with ``cmp -s $old $new || echo "files are different"`` that the supplied binary file is indeed the file provided at https://repo1.maven.org/maven2/com/google/guava/guava/32.1.2-jre.

PR #278, opened 2023-09-11
[#277] Function argument coercion for java interfaces fails when default methods are present

### LINKED ISSUE #277: Argument coercion fails for Java interface with default methods
Same for bound methods.

When checking for suitability (@ PyFunction.java:512 and PyMethod.java:376 as of today) for coercion, the number of methods is check without accounting for default methods.  Affects interfaces like `java.util.function.Consumer<?>` and `java.util.function.BiConsumer<?,?>` where modern java includes a default `.andThen()` method.

It seems to me a simple:
```
	if (method.isDefault()) continue;
```
in the alternate check's loop (ahead of `if (name == null ||...` in each) would let the proxy be created.

The `.invoke()` method of each would then need to pass through calls to default methods, something like:
```
	if (method.getDeclaringClass() == Object.class || method.isDefault()) {
		return method.invoke( this, args );
	} else ...
```

If this seems like a reasonable approach, I'll generate a PR.

pturmel:
Fixes #277

Fix automatic conversions of function/method args for java interfaces with default methods.

Java 8's default methods in interfaces breaks jython's assumptions about lambda-qualified interfaces.  Specifically, jython checks for a single method (or single-named collection of methods) in an interface, but does not exclude interface default methods.

Correct both the decision to supply a proxy in __tojava__() and invocation of the proxy in invoke(), in both PyFunction.java and PyMethod.java.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,16 +7,30 @@ the implementation of Python in Java.
 Only version 2.7 of Python can be supported at present
 (but watch this space for a 3.x version).
 
-Along with good (not perfect!) language
-and runtime compatibility with CPython 2.7,
-Jython 2.7 provides substantial support of the Python ecosystem.
-This includes built-in support of *pip/setuptools*
-(you can use `bin/pip` if the targets do not include `C` extensions)
-and a native launcher for Windows (`bin/jython.exe`)
+## Compatibility
+
+Jython provides good compatibility with Python 2.7 *the language*.
+Also, a high proportion of the standard library is included,
+taken from late versions of CPython (around 2.7.13).
+Some standard library modules have a Jython-specific implementation
+that has not kept pace with its CPython counterpart.
+
+Jython 2.7 support for the Python ecosystem
+includes built-in support of *pip/setuptools*.
+You can use `bin/pip` if the targets do not include `C` extensions.
+There is a native launcher for Windows (`bin/jython.exe`)
 that works essentially like the `python` command.
 
 Jim Baker presented a talk at PyCon 2015 about Jython 2.7,
 including demos of new features: https://www.youtube.com/watch?v=hLm3garVQFo
+
+## Support
+
+Python 2.7 (the language) is no longer supported by the PSF.
+Running on Jython should not be considered an alternative to porting your
+application to Python 3, due to the limitations cited here
+and the small amount of effort available to support 2.7.x.
+Jython 2.7 is offered for continuity because a 3.x is not yet available.
 
 See [ACKNOWLEDGMENTS](ACKNOWLEDGMENTS) for details about Jython's copyright,
 license, contributors, and mailing lists.
@@ -36,6 +50,8 @@ The project uses Git for version-control,
 and the master repository is at https://github.com/jython/jython,
 You should clone this repository to create a buildable copy of the latest state
 of the Jython source.
+Start a new branch for any bug-fix or experimentation you plan.
+
 The previously authoritative repository at https://hg.python.org/jython is not now in use,
 remaining frozen at v2.7.2.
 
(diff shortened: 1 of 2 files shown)


jeff5:
Thanks for offering this. You seem to have a good understanding of the issue. I'll pull this down and play with it.

It needs a set of tests that cover the set of circumstances where this comes up. I also have to ask you to sign the Python CLA (if you haven't already).

pturmel:
Just signed the PSF Contributor agreement, using my GitHub user name.

The interface that blew up on me was `java.util.function.Consumer<T>` due to its `.andThen()` default method.  Many of the interfaces in that package have that or similar problems.  Would a test coercing a callable to `Consumer` suffice?

pturmel:
I suppose it should be noted that this fix breaks java 7 compatibility, if that is a concern.

Stewori:
I personally think that Java 7 compatibility is today not a concern anymore. However, dropping it should be explicitly documented in NEWS and release notes (whenever the release comes).

jeff5:
> I personally think that Java 7 compatibility is today not a concern anymore.

Ceased with 2.7.3
https://github.com/jython/jython/blob/67415e091a1726e66e098a4357dbb22f91420744/NEWS#L56-L57

Now, it is possible to compile with Java 8 and require Java 7 API. I think someone (James maybe) recommended being explicit about this in the build, but I didn't understand how this worked until somewhat later, and perhaps not fully yet. We currently have:
https://github.com/jython/jython/blob/67415e091a1726e66e098a4357dbb22f91420744/build.xml#L94
and
https://github.com/jython/jython/blob/67415e091a1726e66e098a4357dbb22f91420744/build.gradle#L62-L66

Although I noticed just now that our Javadoc link is out of sync with that.
https://github.com/jython/jython/blob/67415e091a1726e66e098a4357dbb22f91420744/build.xml#L1045

jeff5:
> Just signed the PSF Contributor agreement, using my GitHub user name.

Thank-you.

> The interface that blew up on me was `java.util.function.Consumer<T>` due to its `.andThen()` default method. Many of the interfaces in that package have that or similar problems. Would a test coercing a callable to `Consumer` suffice?

The last test in `test/test_java_integration` is `SingleMethodInterfaceTest`. Having looked around existing tests, I think that is the right place for the test and a good template to follow. There we prove that the coercion is effective by actually calling the coerced object. I think it is a good idea to use something ready-made from the Java API and have it work all the way through to a result in context, not just a `__tojava__` call. (You could do that as well, so there's a smoke test to aid debugging.)

Main concern is to have enough variety that we can be sure to notice when someone breaks the implementation.

jeff5:
I notice also you're PR-ing from your `master` branch. Things run much more smoothly if you create a new branch for each contribution. Then you can track our `master` separately and pull from it and as we update `master` .

It may be as simple as renaming your copy, but if it looks risky, we can cope with it as it is. You'll have to make a new fork after we merge this because your `master` will be out of sync completely.

I try to follow the CPython workflow: https://devguide.python.org/getting-started/git-boot-camp/ . There's a Jython developer guide somewhere, based on that, but I just can't keep it up to date. Some adaptation is required `main`->`master`, `python/cpython`->`jython/jython` and so on.

pturmel:
> Things run much more smoothly if you create a new branch for each contribution.

Yeah, I noticed that recommendation *after* I made the PR. ):

I can easily branch this, but can a PR be repointed at it?  (Not new to git, but new to collaboration on GitHub.)

jeff5:
Ok. Probably can't change the PR origin but I'd tolerate a new one if you like. Could try the rename anyway: sometimes GitHub is magic like that.

I'm not new to git either, but I still have to look everything up beyond `pull`, `push `and `status`.

pturmel:
Replacement pull request, using a branch properly, is at #296.

Issue #221, opened 2022-11-30
Incorrect method precedence with overloading and variable arity


tpoliaw:
If a Java class has overloaded methods such as
```java
public class Demo {
    public void foo(int a, int b) {
        System.out.println("two ints");
    }
    public void foo(int... args) {
        System.out.println("vararg ints");
    }
}
```
Calling `demo.foo(1, 2)` from Java code gives `two ints` as expected but from jython code gives `vararg ints`. The same can be seen with constructors now that vararg constructors are supported.

jeff5:
Thanks for finding. Worth a look before there's a 2.7.4.

tpoliaw:
I had a look at this when we initially tried updating but couldn't find a way to fix it that didn't involve duplicating code in several places. I can upload the WIP change as a PR if it'd be any use. Is there any rough timescale of when you're hoping to release 2.7.4?

jeff5:
Having neglected Jython 2 for a while, I'm thinking I should catch up with a 2.7.4. Maybe Christmas?

This has come up in #281 so we ought to try and correct it. Duplicated code doesn't sound like a fatal flaw, but also has a well-known treatment. Let's see the WIP PR. (PRs can be marked as draft.)

tpoliaw:
I'd forgotten I was supposed to be looking at this so thanks for reminding me. I will try and dig out the progress I'd made so far - a few more eyes on it can only be a good thing.

robertpatrick:
@tpoliaw one possible approach would be to change the `matches()` method to return the number of non-varargs arguments matched and have the caller always iterate through the entire list of matching methods/constructors, selecting the one with the highest number of matching arguments.  This is essentially what Java does.

robertpatrick:
What can I do to help?

tpoliaw:
I've opened #282 with my wip from last year. It would be interesting to see if that works and/or whether there are more cases that fall through the gaps. As @jeff5 said, this could really use a thorough set of tests.

robertpatrick:
I will test it and let you know.  I am happy to write tests and actually looked into doing that already.  The thing that was tripping me up was how to create a custom Java class and use it from Python tests.  The current Java test seemed too limiting but I am happy to be convinced otherwise.

tpoliaw:
Have a look at the existing `test_jyoverload.py` and associated `Reflection.java` class as they're are already set up to be accessible. I've already added a few extra cases for overloaded constructors but more can't hurt.

[...]

Issue #281, opened 2023-10-25
Jython 2.7.3 invoking the wrong overloaded Java constructor


robertpatrick:
Our project ([WebLogic Deploy Tooling](https://github.com)) has code that is working properly on Jython 2.7.1.  Our upcoming release has moved to Jython 2.7.3 and the code that was previously working is now failing.  Here is a simplified snippet of what we are doing (with some debugging lines added):

```python
import java.lang.Throwable as Throwable
import oracle.weblogic.deploy.create.CreateException as CreateException

...

def create_create_exception(key, *args, **kwargs):
    """
    Create a CreateException from a message id, list of message parameters and Throwable error.
    :param key: key to the message in resource bundler or the message itself
    :param args: list of parameters for the parameters or empty if none needed for the message
    :param kwargs: contains Throwable or instance if present
    :return: CreateException encapsulating the exception information
    """
    arg_list, error = _return_exception_params(*args, **kwargs)
    print('arg_list (%s) = %s' % (type(arg_list), arg_list))

    arg_len = len(arg_list)
    if error is not None:
        if isinstance(error, Throwable) is False:
            print('error is not Throwable')
            error = convert_error_to_exception()
            print('converted error = %s' % error)
        if arg_len > 0:
            print('Calling CreateException(key, error, arg_list) with error that is an instance of Throwable (%s)' % isinstance(error, Throwable))
            ex = CreateException(key, error, arg_list)
        else:
            print('Calling CreateException(key, error)')
            ex = CreateException(key, error)
    else:
        if arg_len > 0:
            print('Calling CreateException(key, arg_list)')
            ex = CreateException(key, arg_list)
        else:
            print('Calling CreateException(key)')
            ex = CreateException(key)
    return ex
```

The CreateException class has 5 overloaded constructor methods including the two relevant to this situation:

```java
public class CreateException extends BundleAwareException {
    private static final long serialVersionUID = 1L;

    public CreateException() {
        // default constructor
    }

    public CreateException(String messageID) {
        super(messageID);
    }

    public CreateException(String messageID, Object... params) {
        super(messageID, params);
        System.out.println("Inside CreateException(String messageID, Object... params) constructor");
        if (params[0] instanceof Throwable) {
            System.out.println("Inside wrong CreateException() constructor");
        } else {
            System.out.println("Second arg is not a Throwable: " + params[0].getClass().getName());
        }
    }

    public CreateException(String messageID, Throwable cause) {
        super(messageID, cause);
    }

[...]

Issue #276, opened 2023-09-01
Jython 2.7.2 not calling the right overloaded java method for boolean


yogi1967:
I'm using 3rd party jar file which defines:

public class StreamTable extends HashMap {etc}
it defines:

   public final void put(Object key, Object value)
   **public final void put(String key, boolean value)
   public final void put(Object key, boolean value)
   public final void put(Object key, String value)
   public Object put(String key, Object value)
   public final void put(Object key, long value)
   public final void put(String key, long value)
   public final void put(String key, String value)
   public final void put(Object key, int value) 
   public final void put(String key, int value)
 etc

My Jython:
key="a"
value=True
if I call the object using .put(key, value) then it's not calling the .put(String key, boolean value) method. I'm fairly sure it's calling the .put(Object key, Object value) method (or one of the others).

However if I call .put(key, Boolean(value)) then it does call the right boolean method.

Is this a bug?

yogi1967:
... perhaps it's calling the int method.... how does it determine which to call..?

pjs726:
Would it be possible for you to share the jar file or source code? If not, I suppose I could try to recreate the code as you describe it, but there's no guarantee the behavior will be the same.

Also I see you're not sure which method exactly is being called. Would it be possible for you to modify the StreamTable code to give some debug information about which method was called? Something like this:

```
public final void put(Object key, Object value) {
     System.err.println("Object key, Object value");
     ....
}

public final void put(String key, boolean value) {
     System.err.println("String key, boolean value");
     ....
}

....
```

That way we could know for sure which method is being called.

yogi1967:
It's a third-party class and I do not have a java environment to play with.. Here is the decompiled .java file (redacted down to just the methods in question). Hope this is enough... Let me know? Thanks


```
package com.xxx.XXX;

import java.io.Closeable;
import java.io.File;
import java.io.FileInputStream;
import java.io.FileOutputStream;
import java.io.IOException;
import java.io.InputStream;
import java.io.InputStreamReader;
import java.io.OutputStream;
import java.io.OutputStreamWriter;
import java.io.PrintStream;
import java.io.Reader;
import java.io.Serializable;
import java.io.StringReader;
import java.io.StringWriter;
import java.io.Writer;
import java.net.URL;
import java.util.ArrayList;
import java.util.Collection;
import java.util.Collections;
import java.util.Enumeration;
import java.util.HashMap;
import java.util.Iterator;
import java.util.Map;
import java.util.Set;

[...]

PR #282, opened 2023-10-28
Method resolution fix with overloading and variable arity

tpoliaw:
This is very much a WIP and I've not looked at it since I opened #221 last year. I'll try and make sense of what I was doing but open to suggestions of better ways to approach it until then.

### DIFF
--- a/src/org/python/core/PyReflectedConstructor.java
+++ b/src/org/python/core/PyReflectedConstructor.java
@@ -152,25 +152,49 @@ public class PyReflectedConstructor extends PyReflectedFunction {
                 // Remove the keyword args
                 args = new PyObject[allArgs.length - nkeywords];
                 System.arraycopy(allArgs, 0, args, 0, args.length);
+                Object varargMatch = null;
+                ReflectedCallData varargData = null;
 
                 // Look for a constructor with no keyword args
                 for (int i = 0; i < n; i++) {
                     rargs = argslist[i];
                     if (rargs.matches(null, args, Py.NoKeywords, callData)) {
-                        method = rargs.method;
-                        break;
+                        if (!argslist[i].isVarArgs) {
+                            method = argslist[i].method;
+                            break;
+                        } else {
+                            varargMatch = argslist[i].method;
+                            varargData = callData;
+                            callData = new ReflectedCallData();
+                        }
                     }
+                }
+                if (method == null && varargMatch != null) {
+                    method = varargMatch;
+                    callData = varargData;
                 }
             }
        } else {
            // Just look for a constructor with no keyword args
+           Object varargMatch = null;
+           ReflectedCallData varargData = null;
            int n = nargs;
            for (int i = 0; i < n; i++) {
                rargs = argslist[i];
                if (rargs.matches(null, args, Py.NoKeywords, callData)) {
-                   method = rargs.method;
-                   break;
+                    if (!argslist[i].isVarArgs) {
+                        method = argslist[i].method;
+                        break;
+                    } else {
+                        varargMatch = argslist[i].method;
+                        varargData = callData;
+                        callData = new ReflectedCallData();
+                    }
                }
+           }
+           if (method == null && varargMatch != null) {
+               method = varargMatch;
+               callData = varargData;
            }
        }
 
--- a/src/org/python/core/PyReflectedFunction.java
+++ b/src/org/python/core/PyReflectedFunction.java
@@ -162,13 +162,30 @@ public class PyReflectedFunction extends PyObject implements Traverseproc {
     public PyObject __call__(PyObject self, PyObject[] args, String[] keywords) {
         ReflectedCallData callData = new ReflectedCallData();
         ReflectedArgs match = null;
+        // varargs methods should always have lower precedence than a non-varargs
+        // method that also applies but as they will always have <= arguments than an
+        // alternative non-vararg method, they will appear first in the list of options.
+        // Keep track of the first vararg method but don't use it unless no other
+        // method matches.
+        ReflectedArgs varargMatch = null;
+        ReflectedCallData varargData = null;
         for (int i = 0; i < nargs && match == null; i++) {
             if (argslist[i].matches(self, args, keywords, callData)) {
-                match = argslist[i];
+                if (!argslist[i].isVarArgs) {
+                    match = argslist[i];
+                } else {
+                    varargMatch = argslist[i];
+                    varargData = callData;
+                    callData = new ReflectedCallData();
+                }
             }
         }
         if (match == null) {
-            throwError(callData.errArg, args.length, self != null, keywords.length != 0);
+            if (varargMatch == null) {
+                throwError(callData.errArg, args.length, self != null, keywords.length != 0);
+            }
+            match = varargMatch;
+            callData = varargData;
         }
         Object cself = callData.self;
         Method m = (Method)match.method;
--- a/Lib/test/test_joverload.py
+++ b/Lib/test/test_joverload.py
@@ -197,6 +197,23 @@ class VarargsDispatchTests(unittest.TestCase):
 
 
 class ComplexOverloadingTests(unittest.TestCase):
+    def test_constructor_overloading(self):
+        self.assertEqual(Reflection.Overloaded(1).constructorVersion, 'int')
+        self.assertEqual(Reflection.Overloaded(1, 1).constructorVersion, 'int, int')
+        self.assertEqual(Reflection.Overloaded(1, 1, 1).constructorVersion, 'int, int, Object')
+        self.assertEqual(Reflection.Overloaded(1, 1, 1, 1).constructorVersion, 'int, int...')
+
+        self.assertEqual(Reflection.Overloaded(1, [2,3,4]).constructorVersion, 'int, int...')
+
+    def test_method_overloading(self):
+        over = Reflection.Overloaded()
+        self.assertEqual(over.foo(), "int...")
+        self.assertEqual(over.foo(1), "int, int...")
+        self.assertEqual(over.foo(1, 2), "int, int")
+        self.assertEqual(over.foo(1, 2, 3), "int, int, Object")
+        self.assertEqual(over.foo(1, 2, 3, 4), "int, int...")
+
+        self.assertEqual(over.foo(1, [2, 3, 4]), 'int, int...')
 
     def test_complex(self):
         o = Reflection.Overloaded()
--- a/Lib/test/test_joverload.py
+++ b/Lib/test/test_joverload.py
@@ -198,6 +198,7 @@ class VarargsDispatchTests(unittest.TestCase):
 
 class ComplexOverloadingTests(unittest.TestCase):
     def test_constructor_overloading(self):
+        self.assertEqual(Reflection.Overloaded().constructorVersion, '')
         self.assertEqual(Reflection.Overloaded(1).constructorVersion, 'int')
         self.assertEqual(Reflection.Overloaded(1, 1).constructorVersion, 'int, int')
         self.assertEqual(Reflection.Overloaded(1, 1, 1).constructorVersion, 'int, int, Object')
@@ -205,17 +206,17 @@ class ComplexOverloadingTests(unittest.TestCase):
 
         self.assertEqual(Reflection.Overloaded(1, [2,3,4]).constructorVersion, 'int, int...')
 
-        self.assertEqual(Reflection.Overloaded("String", "String", "String").constructorVersion, "string, string, object...")
+        self.assertEqual(Reflection.Overloaded("aa", "bb", "cc").constructorVersion, "String, String, Object...")
 
     def test_method_overloading(self):
         over = Reflection.Overloaded()
         self.assertEqual(over.foo(), "int...")
-        self.assertEqual(over.foo(1), "int, int...")
         self.assertEqual(over.foo(1, 2), "int, int")
         self.assertEqual(over.foo(1, 2, 3), "int, int, Object")
-        self.assertEqual(over.foo(1, 2, 3, 4), "int, int...")
-
         self.assertEqual(over.foo(1, [2, 3, 4]), 'int, int...')
+        # Note in Java these match both foo(int,int...) and foo(int...):
+        self.assertEqual(over.foo(1), "int, int...")
+        self.assertEqual(over.foo(1, 2, 3, 4), "int, int...")
 
     def test_complex(self):
         o = Reflection.Overloaded()
--- a/Lib/test/test_joverload.py
+++ b/Lib/test/test_joverload.py
@@ -138,14 +138,14 @@ class VarargsDispatchTests(unittest.TestCase):
                          "String...:[abc]")
         self.assertEqual(t.test([]),
                          "String...:[]")
-        
+
         self.assertEqual(t.testOneFixedArg("abc"),
                          "String arg1:abc String...:[]");
         self.assertEqual(t.testOneFixedArg("abc", "xyz"),
                          "String arg1:abc String...:[xyz]");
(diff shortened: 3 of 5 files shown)


src/org/python/core/PyReflectedFunction.java
```diff
@@ -162,13 +162,26 @@ protected void addArgs(ReflectedArgs args) {
     public PyObject __call__(PyObject self, PyObject[] args, String[] keywords) {
         ReflectedCallData callData = new ReflectedCallData();
         ReflectedArgs match = null;
+        // varargs methods should always have lower precedence than a non-varargs
+        // method that also applies but as they will always have <= arguments than an
```
tpoliaw, line 166: I'm not sure this is true but I don't think it matters. eg

```python
foo(1,2)
```
would match both
```java
public void foo(int i, int j) {}
public void foo(int i, int j, int... k) {}
```
The vararg case has more arguments but still shouldn't be used. As (I think) the methods are in arg count order, the vararg method will be later and therefore never checked.

tpoliaw:
For reference (mainly to save me having to look it up again next time), [the Java spec reference for overloaded method order](https://docs.oracle.com/javase/specs/jls/se17/html/jls-5.html#jls-5.3).

robertpatrick:
@tpoliaw I cloned your PR branch and tried it.  Unfortunately, it is broken...

```
rpatrick@rpatrick-mac jdk17 % ./jython.sh  
CLASSPATH = /Users/rpatrick/tmp/jdk17/myclasses.jar:/Users/rpatrick/tmp/jython/dist/jython-standalone.jar

  File "/Users/rpatrick/tmp/jdk17/testme.py", line 5, in <module>
    ex = mytest.CreateException("testing CreateException(String message, Throwable cause, Object... params) binding", throwable, ["arg1", "arg2", "arg3"])
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:499)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:480)
	at org.python.core.PyReflectedConstructor.constructProxy(PyReflectedConstructor.java:229)
java.lang.IllegalArgumentException: java.lang.IllegalArgumentException: wrong number of arguments
rpatrick@rpatrick-mac jdk17 % 
```

src/org/python/core/PyReflectedConstructor.java
```diff
@@ -152,26 +152,42 @@ public PyObject __call__(PyObject self, PyObject[] args, String[] keywords) {
...
            for (int i = 0; i < n; i++) {
                rargs = argslist[i];
                if (rargs.matches(null, args, Py.NoKeywords, callData)) {
-                   method = rargs.method;
-                   break;
+                    if (!argslist[i].isVarArgs) {
+                        method = argslist[i].method;
+                        break;
+                    } else if (varargMatch == null) {
+                        varargMatch = argslist[i].method;
```
robertpatrick, line 189: This still isn't correct.  The previous implementation grabs the first method that matches.  This one does exactly the same thing for the vararg methods.  The real problem is that `ReflectedArgs.matches()` only has visibility into a single declared method's parameter definition and the args passed.  It cannot differentiate between multiple overloaded methods (i.e., multiple vararg methods) that match to determine which one is best.  For example, think about this use case:

``` java
mymethod(Object... params)
mymethod(String key, Object... params)
mymethod(String key, Throwable error, Object...params)
```

and then calling it like so:

```python
error = Exception("oops")
obj.mymethod("foo", error, "abc", "def")
```

The problem is that `ReflectedArgs.matches()` will return true for all three of these declared versions of `mymethod()` yet the code picks the first one encountered (when the right one in this particular scenario is actually the one with three params).

The reason that removing #201 works for my overloaded constructors use case is that it does not call `ensureBoxedVarargs()` (because `isVarArgs` is `false`) so it is able to find the correct constructor.

tpoliaw: It may be that the `varargMethod == null` check is causing problems. If it always chose the last matching vararg method it would pick the correct one for your example. Are there alternatives where an earlier one matches better than a later one?

Which would java pick out of

```java
Foo(int i, int... rest) {}
Foo(Object i, Object j, int... rest) {}
Foo(Object i, Object j, Object k) {}
```
When called with `Foo(1, 2, 3)`?

I still don't think this is caused by #201. The same thing can be seen in non-constructors which weren't affected by the change. The reason reverting it fixes this case is that vararg constructors couldn't be called before (without manually passing lists) so they were excluded from the method checks.

tpoliaw:
I'm not surprised, if I thought it was the full fix I'd have submitted it last year. What example were you running to test that, the same one from this issue?

src/org/python/core/PyReflectedConstructor.java
robertpatrick: In Java, it will call `Foo(Object i, Object j, Object k)`...

robertpatrick: @tpoliaw Here is a tweak to your change that I have confirmed resolves the problem for constructors.   The "match score" is really just the length of the varargs method parameters list so this accomplishes finding the best match of the varargs constructors if there is no exact match of a non-varargs constructor.  

I suspect that the `PyReflectedFunction.__call__()` matching logic needs a similar change:

```python
        } else {
            // Just look for a constructor with no keyword args
            List<ReflectedArgs> varargMatches = new ArrayList<>();
            int n = nargs;
            for (int i = 0; i < n; i++) {
                rargs = argslist[i];
                if (rargs.matches(null, args, Py.NoKeywords, callData)) {
                    if (!argslist[i].isVarArgs) {
                        method = argslist[i].method;
                        break;
                    } else {
                        varargMatches.add(argslist[i]);
                    }
                }
            }
            if (method == null) {
                int argsCount = -1;
                for (ReflectedArgs varargsMatch : varargMatches) {
                    if (varargsMatch.args.length > argsCount) {
                        argsCount = varargsMatch.args.length;
                        method = varargsMatch.method;
                    }
                }
            }
        }
```

tpoliaw: And if only the first two options are present it's a compile error so Jython will never have to deal with it.

I think removing the `varargMethod == null` check achieves the same thing. If there is  no matching non-vararg method, the longest (and therefore most specific) vararg method will be used instead.

jeff5:
> For reference (mainly to save me having to look it up again next time), [the Java spec reference for overloaded method order](https://docs.oracle.com/javase/specs/jls/se17/html/jls-5.html#jls-5.3).

And https://docs.oracle.com/javase/specs/jls/se17/html/jls-15.html#jls-15.12.2

src/org/python/core/PyReflectedConstructor.java
robertpatrick: > And if only the first two options are present it's a compile error so Jython will never have to deal with it.
> 
> I think removing the `varargMethod == null` check achieves the same thing. If there is no matching non-vararg method, the longest (and therefore most specific) vararg method will be used instead.

I don’t see how removing the `varargMethod == null` check does the same thing UNLESS the ordering is such that the matching varargs constructor with the longest parameter list is ALWAYS the last one encountered.  Is that a safe assumption to make?  What handles the sorting to ensure that behavior?

robertpatrick: @tpoliaw Of course, you could collapse the logic to check the length of the ReflectedArgs.args to find the longest matching varargs constructor into the first for loop instead of storing references in a List and then iterating again…

tpoliaw:
> I cloned your PR branch and tried it. Unfortunately, it is broken...

I think the reason (or one of them) for this is that `callData` is reused between calls to `arglist[i].matches(...)` and it maintains state which is then passed to the method. This worked for the previous implementation as the matched method would always be the final method checked. Now that we're storing a vararg method we need to cache the autoboxed args somewhere to ensure that the correct arrangement of args are passed to the method when it ends up being called.

src/org/python/core/PyReflectedConstructor.java
tpoliaw: >What handles the sorting to ensure that behavior

It's in `ReflectedArgs.compareTo`, there are other checks in the method for when the args length matches but I think this should ensure the sort order.

```java
        int n = this.args.length;
        if (n < oargs.length) {
            return -1;
        }
        if (n > oargs.length) {
            return +1;
        }
```

tpoliaw:
I'm moving this out of WIP for now. As far as I can make out, this fixes the issues seen with vararg method orders. If the basic premise is ok, I'm happy to look at tidying it up a bit and maybe extracting the common sections.

robertpatrick (commented):
Looks good to me

jeff5:
The test failures on the launcher and spelling are unrelated (and can be fixed by merging from `master` into your branch), but I think the `regrtest` failure is real. Not happening on your machine @tpoliaw ?

tpoliaw:
@jeff5 I must have missed the new one amongst the expected ones. Should be fixed now. Also rebased onto master and squashed a couple of the fix/format changes.

jeff5:
My fix for the Python version in the launcher test has broken again. (Someone is determined to stamp out Python 2: I wonder who?) I'll fix that again.

But `test_joverload` is currently failing too.

tpoliaw:
I'm getting less confident in this PR with every failure. Locally I'm now only getting 

     [exec] 2 tests skipped:
     [exec] 4 tests failed:
     [exec]     test_httplib test_os test_posix test_socket
     [exec] 4 fails unexpected:
     [exec]     test_httplib test_os test_posix test_socket
     [exec] Result: 1

jeff5:
> I'm getting less confident in this PR with every failure. Locally I'm now only getting
> 
> ```
>  [exec] 2 tests skipped:
>  [exec]     test_codecmaps_hk test_curses
>  [exec] 4 tests failed:
>  [exec]     test_httplib test_os test_posix test_socket
>  [exec] 4 fails unexpected:
>  [exec]     test_httplib test_os test_posix test_socket
>  [exec] Result: 1
> ```
I would run them individually to see the manner they fail, or cut and paste the list of tests from that output to a command like:
```
dist/bin/jython -m test.regrtest -v test_httplib test_os test_posix test_socket
```
If the complaints are about numbers of arguments, we're probably not there yet. If it's networking, they may be spurious (e.g. firewall or ISP interference).

tpoliaw:
There's nothing that looks related and I'm getting all the same errors when building `master` instead of this branch

jeff5:
Well, it passed here on GitHub, and for me at home, so I'll dive into the code with the questions in my mind being:

- [x] Do the test answers reflect the behaviour we want?
- [x] Is there enough cover?
- [x] Is the style ok?

If we have to incorporate from ` master` again, the way I do it is: `get merge master` on the PR branch. Force pushing invalidates what others may have pulled. It looks rambling but it all get's squeezed out in the squash merge.

tpoliaw:
Thanks for looking through it. The main thing I'm not happy about is the duplication and whether it can be extracted to a single place. 

>If we have to incorporate from  master again, the way I do it is: get merge master on the PR branch. Force pushing invalidates what others may have pulled. It looks rambling but it all get's squeezed out in the squash merge.

Sorry about that, it's habit from the Gerrit workflow where the commit is reviewed unit rather than change as a whole.

jeff5:
The main question, of course, is "what would Java do". The [specification](https://docs.oracle.com/javase/specs/jls/se17/html/jls-15.html#jls-15.12) is quite long and it prescribes several steps and sub-steps (phases) for finding the right definition. It covers a lot of possibilities even when we ignore things Jython 2 doesn't understand (type variables, lambdas). I'm fairly sure our process is not the one prescribed, but do we get the same result?

Since I don't hold all this in my head, I've added to `javatests.Reflection` a `main()` that calls the same methods from Java. First interesting observation is that these in `Reflection.Overloaded` don't compile:
```
            foo(1);
            foo(1, 2, 3, 4);
```
because they match both `foo(int, int...)` and `foo(int...)` whereas Jython selects the latter.

I've been less successful showing we respect the rule about selecting the most specific definition. If I introduce
```
        public String foo(Number a, int... b) { return "Number, int...";   }
        public String foo(Object a, int... b) { return "Object, int..."; }
```
It makes different choices from Java when the first argument s `1L`. I think this is because we only ask whether a conversion is possible. We deal with specificity by ordering the signatures, but it looks like black magic to me at the moment.

Combine this with the fact that `PyLong.__tojava__` performs narrowing conversions to `int` etc., which seems dubious to me, and `int` is as good a match for `1L` as `Number`.

However, I bet a load of stuff will break if we change that.

robertpatrick:
A couple of thoughts:

- yes, foo(int a, int b) and foo(int a, int… args) is ambiguous.  The real question is do we have to catch this in Jython during method selection or is it ok to let Java catch this for us?  Since this won’t even compile, I think it is perfectly fine to not figure out that it is ambiguous in Jython because Java will never actually allow it to happen.  Even if something were to slip through, I think it is perfectly fine to have the JVM generate the error at dispatch time rather than Jython needing to catch it.
- Jython has long suffered from issues with converting literal numbers that could/should be ints to PyLong (doesn’t Python 3 do away with different integer types anyway?).  Again, I don’t think we need to try to fix that as part of this fix.  Just change your overload test cases to avoid int vs. long literals in Python (e. g., use Integer and Long Java classes instead).  Better yet, just change the test cases to use other types to test overloaded method selection.

Jython may never be perfect but last I checked, this PR fixes a real problem and at least, does as good a job as (in fact, a better job than) Jython 2.7.1 and before.  My vote would be to wrap this PR up, merge it, and let others report issues if they find them.  Clearly, the current 2.7.2/2.7.3 is badly broken and this change supports our real-world use cases as far as I have been able to test it so while it is maybe not perfect, it at least restores pre-2.7.2 functionality that was broken in 2.7.2/2.7.3.

jeff5:
> * Since this won’t even compile, I think it is perfectly fine

That text (as Java, in context) won't compile, to save you from the ambiguity, but it doesn't stop Jython interpreting the similar expression in Python as unambiguously a call to just one of those definitions.

Python 3 has only one kind of `int`. That may cause more problems than it solves, as noted in https://github.com/jython/jython/issues/281#issuecomment-1784101569. (Not a problem for now.)

robertpatrick:
> > * Since this won’t even compile, I think it is perfectly fine
> 
> That text (as Java, in context) won't compile, to save you from the ambiguity, but it doesn't stop Jython interpreting the similar expression in Python as unambiguously a call to just one of those definitions.

again, the JVM will never allow the call (to the wrongly selected method) to even occur, whether due to a compiler error or a runtime dispatch error.

> Python 3 has only one kind of `int`. That may cause more problems than it solves, as noted in [#281 (comment)](https://github.com/jython/jython/issues/281#issuecomment-1784101569). (Not a problem for now.)

So where does that leave us?  What do we need to do to get this fix merged (and 2.7.4 released)?

jeff5:
> So where does that leave us?

You and Peter @tpoliaw have dug into this fairly thoroughly (thank you) and I've played around with it enough now to follow your discussion. If you're happy this fixes the variable arity call I am too. 

It doesn't implement the standard process exactly and can't without changes that would almost certainly be problematic elsewhere. (Worth re-thinking for Python 3, though.) So I expect more problems. In fact, I think #276 is one.

robertpatrick:
@jeff5 I think the issue #276 has nothing to do with how the method selection algorithm is working and more to do with the types of the Python int/long args by the time they get to the method selection code.  Should be easy enough to verify…

PR #286, opened 2023-11-15
Cleanup of obsolete Mercurial files

meenbeese:
I removed several unused files dating back to the Mercurial to Git migration:
- .hgignore
- .hgsub
- .hgsubstate


jeff5:
I think you're right. I see no reason to keep these.

jeff5:
Thanks for contributing this @meenbeese .

PR #296, opened 2023-12-31
Fix automatic conversions of function/method args for java interfaces…

### LINKED ISSUE #277: Argument coercion fails for Java interface with default methods

pturmel:
Fixes #277

… with default methods.

Java 8's default methods in interfaces breaks jython's assumptions about lambda-qualified interfaces.  Specifically, jython checks for a single method (or single-named collection of methods) in an interface, but does not exclude interface default methods.

Correct both the decision to supply a proxy in __tojava__() and invocation of the proxy in invoke(), in both PyFunction.java and PyMethod.java.


pturmel:
Prior discussion on #278.

Test case to follow.

pturmel:
Test cases complete.

pturmel:
I couldn't help but notice that the Callable coercion test is disabled.  I investigated, because I thought I'd add a test case for Callable conversion to Consumer, too. (Did not work, of course.)  I suspect the balance of that old issue could be solved with similar proxy code in `PyInstance`.

(I did not find that old issue in the github list.  Should it be resurrected for a possible complete fix?)

jeff5:
> I couldn't help but notice that the Callable coercion test is disabled. ... (I did not find that old issue in the github list. Should it be resurrected for a possible complete fix?)

I looked at that too. It sounds like Jim had an idea but it wasn't as simple as he thought. If you think you have an answer feel free to replicate the issue on GH. Unlike CPython, we have not replicated open BJO issues mechanically, and anyway, I'd appreciate it explained more carefully than Jim has there.

The idea seems to be that an object with `__call__` should automatically be a Java `Callable`. However, note that `PyInstance` is only for [classic (old-style) classes](https://docs.python.org/2.7/reference/datamodel.html#new-style-and-classic-classes).

Separate PR if you think you have a  fix.

pturmel:
> only for [classic (old-style) classes](https://docs.python.org/2.7/reference/datamodel.html#new-style-and-classic-classes).

Ew.  Ok, not so simple.  I'll poke at it later.

Lib/test/test_java_integration.py
```diff
@@ -958,6 +965,19 @@ def test_some_noncallable_object(self):
...
+            x.append(item)
+        future = self.executor.submit(ConsumerCaller(f))
+        future.get()
+        self.assertEqual(x, [42])
+
+    def test_consumer_method(self):
+        obj = ConsumerMethod()
+        future = self.executor.submit(ConsumerCaller(obj.doit))
+        future.get()
+        self.assertEqual(obj.data, [42])
```
jeff5, line 981: Two lines around top-level functions and classes.

src/org/python/core/PyFunction.java
```diff
@@ -518,6 +518,8 @@ public Object __tojava__(Class<?> c) {
                 String name = null;
                 for (Method method : c.getMethods()) {
                     if (method.getDeclaringClass() != Object.class) {
+                        if (method.isDefault())
```
jeff5: We have braces around all bodies (in any code less than about 15 years old). For ease of reading, I would do this as:
```
if (...) {
    continue;
} else if (...) {
    ...
} else {
    ...
}
```

src/org/python/core/PyMethod.java
```diff
@@ -382,6 +382,8 @@ public Object __tojava__(Class<?> c) {
                 String name = null;
                 for (Method method : c.getMethods()) {
                     if (method.getDeclaringClass() != Object.class) {
+                        if (method.isDefault())
```
jeff5: And the same `if (...) { ... } else if (...)` here.

jeff5 (changes requested):
This looks good to me. I took the time to single-step through a lot of it, most of which you haven't changed, because it's quite intense stuff and I wouldn't say my grasp of it was perfect. It was more a question of feeling comfortable this is the behaviour we want, rather than questioning whether it is done right. The tests, though simple, helped a lot thanks.

There are minor formatting problems in the new material, but that's all.

Also, would you please:

- [ ] Merge from `master` into your PR branch.
- [ ] Add a line to NEWS like `[ GH-277 ] ...` 
- [ ] Add yourself to ACKNOWLEGMENTS.
- [ ] Run the regression tests again.
- [ ] Commit and push.

Thanks for doing this.

pturmel:
Done, but with a slight tweak.

I added the braces around `continue;` as requested, but omitted the following `else`.  With it, IntelliJ flagged the `continue` as unnecessary, but it really is necessary. Any code added in the future after the next `if` would be accidentally run if a code cleanup deleted the `continue` as IntelliJ was recommending.

jeff5 (approved):

jeff5:
I see where IntelliJ is coming from. Since there is no code after the if-nest, the body could be empty. However, that would look odd. What you have is fine (conformant).

My personal preference is for if statements to leave by the bottom (maybe `return` in simple code), so I'm not fond of `continue`. Here it would would probably be `// pass`.

jeff5:
Thanks for this @pturmel, it was an education to review it.

pturmel:
Heh. Jython has been cracking my skull for 15 years or so, thanks to a particular commercial user of jython with whom I'm heavily involved.  (Immediately obvious if you google me.)  Turnabout is fair play, after so many years. 😁

PR #303, opened 2024-02-07
Remove unused LineNumberTable

### LINKED ISSUE #280: LineNumberTable deprecated for removal in 2.6 still there
I just randomly noticed that org.python.compiler.LineNumberTable contains the deprecation note:
"Not used. Delete in 2.6."

If it is still in use, the note should be removed, otherwise the class. This is not high priority, this issue is opened mainly as a reminder.

tpoliaw:
Saw this in the 2.7.4 milestone list and it looked like a quick fix.

Fixes #280


jeff5:
Thank you! It looks a safe change. (Just running tests.)

PR #325, opened 2024-04-09
Add upward compatibility to Java 9 Modularity(2)

R077A6r1an:
I've created the new branch with the changes discussed at https://github.com/jython/jython/pull/323 and signed the Python contributor agreement.

### DIFF
--- a/ACKNOWLEDGMENTS
+++ b/ACKNOWLEDGMENTS
@@ -205,6 +205,7 @@ In no particular order they are:
     Diogo Teles Sant'Anna (GitHub: diogoteles08)    C, 2023-08-10
     Phil Turmel (GitHub: pturmel)                   C, 2023-12-30
     lifebarier (GitHub: lifebarier)                 C, 2024-03-06
+    Adrian Ross (GitHub: R077A6r1an)                C, 2024-04-09
 
 
 Since about 2012 we have been asking contributors to sign the Python
--- a/NEWS
+++ b/NEWS
@@ -16,14 +16,24 @@ Jython 2.7.4
 
 New Features
 
-    - When using Java 9+ modules, a Jython jar on the module path now appears
-      as an automatic module org.python.jython2, org.python.jython2.standalone,
-      or org.python.jython2.slim, according to the JAR [GH-325]. These modules
-      overlap in content and may not be used together.
+    - When building a modular application or library, the jython-slim JAR may
+      be placed on the module path [GH-325]. It will export its usual packages
+      in an automatic module org.python.jython2. Application build tools that
+      create a run script, will generally add the dependencies of Jython to
+      the module path automatically. The developer may have to add (via the
+      build script) --add-modules ALL-MODULE-PATH to the java command to
+      ensure that these dependencies are resolved.
 
-Jython 2.7.4b2 Bugs fixed
+    - The jython-standalone JAR (module org.python.jython2.standalone) may
+      also be used. Many shaded (and some unshaded) dependencies are then
+      duplicated in its module, which can be a problem. It may be necessary to
+      --add-modules ALL-DEFAULT to ensure JDK classes are resolved.
+
+
+Jython 2.7.4b2 Bugs fixed and features
 
     - [ GH-325 ] Upward compatibility to Java 9 Modularity
+
 
 Jython 2.7.4b1 Bugs fixed
     - [ GH-316 ] Clash over META-INF in shaded JARs
--- a/NEWS
+++ b/NEWS
@@ -16,8 +16,10 @@ Jython 2.7.4
 
 New Features
 
-    - In a module Java project, Jython now appears as the automatic module
-      org.python.jython (whichever JAR you use) thanks to GH-325.
+    - When using Java 9+ modules, a Jython jar on the module path now appears
+      as an automatic module org.python.jython2, org.python.jython2.standalone,
+      or org.python.jython2.slim, according to the JAR [GH-325]. These modules
+      overlap in content and may not be used together.
 
 Jython 2.7.4b2 Bugs fixed
 
--- a/NEWS
+++ b/NEWS
@@ -16,10 +16,12 @@ Jython 2.7.4
 
 New Features
 
-    - Bug fixes only.
+    - In a module Java project, Jython now appears as the automatic module
+      org.python.jython (whichever JAR you use) thanks to GH-325.
 
 Jython 2.7.4b2 Bugs fixed
 
+    - [ GH-325 ] Upward compatibility to Java 9 Modularity
 
 Jython 2.7.4b1 Bugs fixed
     - [ GH-316 ] Clash over META-INF in shaded JARs
--- a/build.gradle
+++ b/build.gradle
@@ -649,6 +649,7 @@ jar {
         attributes ([
             'Main-Class': 'org.python.util.jython',
             'Built-By': 'build.gradle',
+            'Automatic-Module-Name': 'org.python.jython'
         ])
 
         attributes( [ // Build-Info section
--- a/build.gradle
+++ b/build.gradle
@@ -649,7 +649,7 @@ jar {
         attributes ([
             'Main-Class': 'org.python.util.jython',
             'Built-By': 'build.gradle',
-            'Automatic-Module-Name': 'org.python.jython2.slim'
+            'Automatic-Module-Name': 'org.python.jython2'
         ])
 
         attributes( [ // Build-Info section
--- a/build.gradle
+++ b/build.gradle
@@ -649,7 +649,7 @@ jar {
         attributes ([
             'Main-Class': 'org.python.util.jython',
             'Built-By': 'build.gradle',
-            'Automatic-Module-Name': 'org.python.jython'
+            'Automatic-Module-Name': 'org.python.jython2.slim'
         ])
 
         attributes( [ // Build-Info section
--- a/build.xml
+++ b/build.xml
@@ -939,6 +939,7 @@ The text for an official release would continue like ...
             <manifest>
                 <attribute name="Main-Class" value="org.python.util.jython" />
                 <attribute name="Built-By" value="${user.name}" />
+                <attribute name="Automatic-Module-Name" value="org.python.jython"/>
                 <attribute name="Implementation-Vendor" value="Python Software Foundation"/>
                 <attribute name="Implementation-Title" value="Jython fat jar with stdlib"/>
                 <attribute name="Implementation-Version" value="${jython.version}"/>
@@ -982,6 +983,7 @@ The text for an official release would continue like ...
             <manifest>
                 <attribute name="Main-Class" value="org.python.util.jython" />
                 <attribute name="Built-By" value="${user.name}" />
+                <attribute name="Automatic-Module-Name" value="org.python.jython"/>
                 <attribute name="Implementation-Vendor" value="Python Software Foundation"/>
                 <attribute name="Implementation-Title" value="Jython"/>
                 <attribute name="Implementation-Version" value="${jython.version}"/>
@@ -1011,6 +1013,7 @@ The text for an official release would continue like ...
             <manifest>
                 <attribute name="Main-Class" value="org.python.util.jython" />
                 <attribute name="Built-By" value="${user.name}" />
+                <attribute name="Automatic-Module-Name" value="org.python.jython"/>
                 <attribute name="Implementation-Vendor" value="Python Software Foundation"/>
                 <attribute name="Implementation-Title" value="Jython Test JAR"/>
                 <attribute name="Implementation-Version" value="${jython.version}"/>
--- a/build.xml
+++ b/build.xml
@@ -916,6 +916,7 @@ The text for an official release would continue like ...
             <manifest>
                 <attribute name="Main-Class" value="org.python.util.jython" />
                 <attribute name="Built-By" value="${user.name}" />
+                <attribute name="Automatic-Module-Name" value="org.python.jython2"/>
                 <attribute name="Implementation-Vendor" value="Python Software Foundation"/>
                 <attribute name="Implementation-Title" value="Jython fat jar"/>
                 <attribute name="Implementation-Version" value="${jython.version}"/>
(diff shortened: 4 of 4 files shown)


jeff5:
Looks good to me. I'll just add it to NEWS. I think it counts as a feature.

jeff5:
**On second thoughts**, reading: https://dev.java/learn/modules/automatic-module/#automatic-module-names---small-detail-big-impact and the section "When to Set Automatic-Module-Name" there's a caution to give module names a stable meaning. It's what I worried might be the case in https://github.com/jython/jython/pull/323#issuecomment-2041449495, that the same named module had different content.

I tried two JARs on the module path supplying the same module and it didn't cause a problem. (`java` takes the first.)

I can see that a project requiring a specific module might find it problematic if the module content was different depending what JAR was on the module path. If an application looks for a class that is in one but not the other (because it looks for the shaded name when the unshaded is first, or _vice versa_) it will not be found. That inclines us to the clarity of plain, `.standalone` and `.slim` module names.

It is also problematic if different modules supply the same package, as they will if you put two of our JARs on the module path. But that's a noisy failure at build time.

I've marked this a draft (don't merge) for now, while I'll try this locally, to check my understanding.

R077A6r1an:
Okay, understandable. I self will also have a look at it and share later my results.

jeff5:
That was worth doing (even if we change our minds) because I spotted an omission. We were not giving an explicit module name to the JAR that results from installing Jython. :/
```
PS 325> java --module-path "inst-old\jython.jar" org.python.util.jython -c "import org.python; print(org.python.Version.getModule())"
module jython
```
With a correction I get:
```
PS 325> java --module-path "inst\jython.jar" org.python.util.jython -c "import org.python; print(org.python.Version.getModule())"
module org.python.jython2
```

I'm going for the `jython2` name at the moment because of what the tutorial said about stable module design. It can't mean frozen entirely or there could be no innovation. It probably means additions only, or a long deprecation, as we would/should with JARs now. However, Jython 3 will be a lot different (and less visible). I don't want to promise anything.

A clash of JARs gives a noisy error:
```
PS 325> java --module-path "inst\jython.jar;kit\jython-standalone.jar" org.python.util.jython -c "import org.python; print(org.python.Version.getModule())"
Error occurred during initialization of boot layer
java.lang.module.ResolutionException: Module org.python.jython2.standalone contains package org.python.apache.commons.io.input, module org.python.jython2 exports
package org.python.apache.commons.io.input to org.python.jython2.standalone
```
I think this is preferable to a surprise missing class later when the `org.python.jython` module you've got (because it appeared first on the path) turns out different from the one you designed for.

A conflicting JAR on the class path gives no error, because them's the rules.
```
PS 325> java --module-path "inst\jython.jar" --class-path "kit\jython-standalone.jar" org.python.util.jython -c "import org.python; print(org.python.Version.getModule())"
module org.python.jython2
```

I'll push this to the PR branch so you can see, then I'll hammer on it locally as the CI doesn't test a built Jython.

R077A6r1an:
Okay, in the meantime, I was testing and thinking of that the issue with the different content in the jar's does not only apply to the modularization. Actually it does also affects the current build. Changing the classpath order for the distributions Jython Slim and Jython normal, so that Jython normal is found first and loaded, the `Lib` folder is not included, and JVM startup fails with

```
Exception in thread "main" ImportError: Cannot import site module and its dependencies: No module named site
Determine if the following attributes are correct:
  * sys.path: [/Users/adrian/Documents/xcode/docker/jython/test_suite/lib/Lib, __classpath__, __pyclasspath__/]
    This attribute might be including the wrong directories, such as from CPython
  * sys.prefix: /Users/adrian/Documents/xcode/docker/jython/test_suite/lib
    This attribute is set by the system property python.home, although it can
    be often automatically determined by the location of the Jython jar file

You can use the -S option or python.import.site=false to not import the site module
```

Now on modular level, it may be solved with setting different module names for each distribution (except for the split packages). But for non-modular users right now, having both on class path results in the same exception as above (maybe not directly, but maybe of any reason caused through manifest class path setting or manually set class path on cmdline in the false order.

jeff5:
What did you launch? I don't get that problem locally.
```
PS 325> java --class-path "inst\jython.jar;kit\jython-standalone.jar" org.python.util.jython
Jython 2.7.4b2-SNAPSHOT (heads/pr-325-modules-dirty:544614bdb, Apr 15 2024, 17:03:58)
[Java HotSpot(TM) 64-Bit Server VM (Oracle Corporation)] on java11.0.22
>>> import sys
>>> sys.path
['', 'C:\\Users\\Jeff\\Documents\\Jython\\gh-iss\\325\\inst\\Lib', '__classpath__', '__pyclasspath__/', 'C:\\Users\\Jeff\\.local\\lib\\jython2.7\\site-packages', 'C:\\Users\\Jeff\\Documents\\Jython\\gh-iss\\325\\inst\\Lib\\site-packages']
```

R077A6r1an:
I've tested with using Jython as a library.  The source code was just

```java
package test;

import org.python.util.PythonInterpreter;

public class Main {
  public static void main(String[] args) {
    PythonInterpreter interpreter = new PythonInterpreter();
    interpreter.exec("from java.lang import System \nSystem.out.println(\"Module test\")");
  }
}
```

and building and testing with script

```bash
#!/bin/sh -e

cd src;
javac -d ../out test/Main.java --module-path ../lib/jython-dev.jar
cd ../out
jar --create --file module-test.jar --manifest Manifest.txt test/Main.class

# Now test
java -cp ../lib/jython.jar -jar module-test.jar
```

and the custom Manifest is

```
Manifest-Version: 1.0
Created-By: JAR
Class-Path: ../lib/jython.jar ../lib/jython-standalone.jar ../lib/jython-dev.jar
Main-Class: test.Main
```

My working directory is

```
.
├── compile.sh
├── lib
│   ├── jython-dev.jar
│   ├── jython-installer.jar
│   ├── jython-standalone.jar
│   └── jython.jar
│ ── out
│   ├── Manifest.txt
│   ├── module-test.jar
│   └── test
│       └── Main.class
└── src
    └── test
        └── Main.java
```

and the jars at `lib` are the shaded artifacts build with ant.

jeff5:
These are not circumstances any normal user would encounter. It's hard to guess where Jython would look for `site-packages`, `./Lib` probably.

We don't distribute `jython-dev`. It is only used in the development environment, where we can avoid compiling and JARring up the standard library. If Jython is run from `jython-dev`, it will assume a layout like that of an installed Jython (and no shading).

jeff5:
JARs made this way work for me when used normally to run regression tests.

I haven't been able to make a genuine modular application that requires `jython.slim`. I've tried this with Gradle, and I get a launch script with both a module path and a class path that look plausible. Running it, however, throws a `NoClassDefFoundError: org/objectweb/asm/ClassVisitor`.

The application is the Java-wrapped version of `import org.python; print org.python.Version.getModule()` as above. The non-modular version of the application reported the unnamed module.

This may be telling me something useful about a bug in our the dependencies that I've missed up to now. Or it may just be another difficult to get right aspect of using Gradle.

R077A6r1an:
Okay, now I've tried with an modular application in maven with the new automatic module name set, running your code `import org.python; print org.python.Version.getModule()` with the `PythonInterpreter` class and output is

```
unnamed module @34710b93
```

So might be a Gradle thing. Don't know exactly, I'm more working with maven.

jeff5:
Shouldn't that report the module name we chose for the module that contains `org.python`?

I get that kind of result with Gradle when I *don't* try to make a modular application. Failures when I do, I put down to not understanding Gradle well enough yet.

R077A6r1an:
Now I’m not sure. In the python interpreter it prints that. But I will test what pure Java says about that. It may be that for Java (I‘m using openjdk21), a automatic module name is only for the name for requiring and rest is interprered as unanmed module, or similar. But I will have a look over that in both modular and non-modular application.

Now about gradle I don‘t know why it fails. In maven I have used a hacky way to repackage at build time the latest jython-slim from central with the automatic module name attribute set, and then in a submodule using it as dependency, with modified compiler arguments to use the manipulated jar. Now it may be that the not manipulated jython is taken by executing, but I don‘t think so. I‘ve excluded it from the build, but I will have a look on that, and test aswell the application with the jython-slim artifact build with ant.

R077A6r1an:
Now I think it apears as unnamed because I‘ve accidently used the artifact on class path. I will change it to module path when I test it later.

R077A6r1an:
Okay, thanks. Was my mistake with  having jython on the classpath instead of on the module path. When using it on module path, then output is

```
module org.python.jython2.slim
```

as you expected in both, modular and non-modular application.

jeff5:
Hurrah! I can get that too now, but only by requiring `org.objectweb.asm` in my application and the artefact (JAR) in `build.gradle`, for which it has no direct need. If I don't, running it gives me:
```
Exception in thread "main" java.lang.NoClassDefFoundError: org/objectweb/asm/ClassVisitor
        at org.python.jython2.slim@2.7.4b2-SNAPSHOT/org.python.core.PySystemState.initPackages(PySystemState.java:1316)
        at org.python.jython2.slim@2.7.4b2-SNAPSHOT/org.python.core.PySystemState.doInitialize(PySystemState.java:1241)
        at org.python.jython2.slim@2.7.4b2-SNAPSHOT/org.python.core.PySystemState.initialize(PySystemState.java:1158)
...
```
This makes me think `jython-slim` may not be specifying its dependencies correctly -- a good thing to have found out if true.

Gradle constructs the same module and class paths (except for order) whether I add this dependency to my application or not. In particular, `asm-9.6.jar` is on the module path in both cases, but it is not mentioned in the module graph if I don't.

Did you have to make any such special provision in your Maven build?

R077A6r1an:
No, in my mavan build I‘ve just required jython slim as normaly, and with the asm library I‘ve touched in fact nothing, except for that it is on my module path. Same works if on class path.

jeff5:
I think I understand this now.

It says [here](https://dev.java/learn/modules/automatic-module/#automatic-modules) "an automatic module reads all other modules that make it into the graph". With Jython as an automatic module, required by my modular project, I observe that all the other *automatic* modules are in the graph, as it says [here](https://dev.java/learn/modules/automatic-module/#module-resolution-for-automatic-modules): "the module system pulls in _all_ automatic modules once it encounters the _first one_ that is explicitly required".

In fact, I _only_ get the automatic modules. Those JARs in the module path that contain explicit modules (`module-info.class`) are not in the module graph.

Gradle knows what JARs we need (from the POM, I assume) and puts them either on the module path or on the class path. If they are modules, because of either a `module-info.class` or `Automatic-Module-Name`, the JARs have a place on the module path. Plain JARs without either go on the class path.

The next bit was surprising. The `asm` JARs have a `module-info.class` and are on the module path, but not in the module graph because nothing _explicitly_ requires them. Unfortunately, we need them for Jython to work, so the application built with Gradle fails.

The allocation of `asm` to the module path seems logical to me, and the behaviour that ignores it is in the JVM, not Gradle. One has to `--add-modules ALL-MODULE-PATH` to get the desired output `module org.python.jython2.slim`.

jeff5:
I think this is ok now, at least as far as use of `jython-slim` goes. I'll document in NEWS that using this feature requires `--add-modules ALL-MODULE-PATH` be specified to the `java` command.

I also got `jython-standalone` to work in a modular app, but this time I needed `--add-modules=ALL-SYSTEM`.

I'm not sure one would want to use the other JARs in a modular project. How do you intend to use Jython as a module @R077A6r1an ?

R077A6r1an:
> I think I understand this now.
> 
> It says [here](https://dev.java/learn/modules/automatic-module/#automatic-modules) "an automatic module reads all other modules that make it into the graph". With Jython as an automatic module, required by my modular project, I observe that all the other _automatic_ modules are in the graph, as it says [here](https://dev.java/learn/modules/automatic-module/#module-resolution-for-automatic-modules): "the module system pulls in _all_ automatic modules once it encounters the _first one_ that is explicitly required".
> 
> In fact, I _only_ get the automatic modules. Those JARs in the module path that contain explicit modules (`module-info.class`) are not in the module graph.
> 
> Gradle knows what JARs we need (from the POM, I assume) and puts them either on the module path or on the class path. If they are modules, because of either a `module-info.class` or `Automatic-Module-Name`, the JARs have a place on the module path. Plain JARs without either go on the class path.
> 
> The next bit was surprising. The `asm` JARs have a `module-info.class` and are on the module path, but not in the module graph because nothing _explicitly_ requires them. Unfortunately, we need them for Jython to work, so the application built with Gradle fails.
> 
> The allocation of `asm` to the module path seems logical to me, and the behaviour that ignores it is in the JVM, not Gradle. One has to `--add-modules ALL-MODULE-PATH` to get the desired output `module org.python.jython2.slim`.

Yes, I've just kept testing and found out that this also occures to me when running all completly on module path. I thought I would run all on module path but I've just noticed that the classpath was set automaticly by the generated manifest. So think that's okay.

I've also just tested if same occures if using instead of `jython-slim` `jython-standalone`, but seams this is only issue of `jython-slim`. As I've noticed that the downloaded `jython-standalone` jar results in a fat 50MB package with no dependencies neither local, nor listed on maven central and Jython normal, I think that this are the shaded artifacts generated from ant. So in this case they must have registered it's packages as part of their module and that does not cause the issue with `asm` (but this may cause split-package issues when declaring the included dependencies as project dependency).

Now in my case, I'm using `jython-slim` for server-side python scripting and plugin capabilities with interoperability with Java. As like the WIP server uses a modular architecture to just run with the modules within all required features, I'm working modular to declare all internal server-module dependencies.

jeff5:
I'm glad you intend to use `jython-slim`. While alert to the possibility we're doing that JAR incorrectly (it's my addition) I think that's preferable in a modular project. For this reason, I'm using the plain name `org.python.jython2` for that module and longer names for the others. I can get jython-standalone to work in a modular build, but I don't want to encourage it.

I've updated the NEWS entry again to reflect that and what I've learned about where `--add-modules` is needed.

jeff5:
Failing test on MacOS is to do with GitHub advancing the OS version from a 12 to a 14 image on which Java 8 has not been installed. (Can ignore for now: need to ask for Java 11 or pin at MacOS 12.)

R077A6r1an:
Okay, that seems right. Other than that with the `asm` library in a modular project, I don't find any other issues. Theoretically it could be solved with a module info, but actually that would break even more considering that some dependencies, specially antlr runtime, don't even have an automatic module name.

jeff5:
> Okay, that seems right. Other than that with the `asm` library in a modular project, I don't find any other issues. ...

Thanks for the second pair of eyes. I don't regard this as an issue, except with my own understanding that it doesn't "just work". Fair to say I have learned a lot about the module system in the last few weeks while doing this. You will see I have written as succinctly as I can in NEWS what options are necessary. It should all appear under your name, however.

No, I don't want to add a `module.info` in Jython 2, but what I have learned informs future ideas.

R077A6r1an:
Nice to hear that. Thanks for your time testing and merging my changes.

jeff5:
[A beta 2 is now out](https://central.sonatype.com/namespace/org.python) with this change (and a fix for the failing launcher test).

Issue #42, opened 2020-07-25
Handle IndexError from embedded interpreter (PR 42)


jeff5:
https://github.com/jython/jython-bad-history/pull/42

jhubbardnso:
https://github.com/jython/jython/pull/331 has been opened to restore this fix.

PR #331, opened 2024-06-05
Catch IndexError when accessing sys.argv

jhubbardnso:
This change was approved as part of https://github.com/jython/jython-bad-history/pull/42 but it appears that it was submitted during the period where the repo had a broken history.   As such and I'm submitting it again for consideration.  

As before, our application replies on an embedded interpreter and when run without this change we see:

```$ echo "print(Misc.getRelease())" | RunScript
Exception in thread "main" Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/jhubbard/Downloads/jython-standalone-2.7.3.jar/Lib/warnings$py.class", line 226, in warn
  File "/home/jhubbard/Downloads/jython-standalone-2.7.3.jar/Lib/warnings$py.class", line 226, in warn
IndexError: index out of range: 0
```
The fix here is to catch the index error in the except block and assume `__main__` just like in the event of a Attribute or Type Error.

### DIFF
--- a/ACKNOWLEDGMENTS
+++ b/ACKNOWLEDGMENTS
@@ -206,6 +206,7 @@ In no particular order they are:
     Phil Turmel (GitHub: pturmel)                   C, 2023-12-30
     lifebarier (GitHub: lifebarier)                 C, 2024-03-06
     Adrian Ross (GitHub: R077A6r1an)                C, 2024-04-09
+    John Hubbard (GitHub: jhubbardbnso)             C, 2024-06-06
 
 
 Since about 2012 we have been asking contributors to sign the Python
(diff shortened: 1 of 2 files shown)


jhubbardnso:
This addresses https://github.com/jython/jython/issues/42.

Stewori:
I think this is okay (the codespell issue is unrelated to this PR) and it can be merged. @jhubbardnso Have you signed the contributor agreement? If not, please sign it and add yourself to https://github.com/jython/jython/blob/master/ACKNOWLEDGMENTS with "C, [date you signed the agreement]". Thanks!

jhubbardnso:
I just signed the the contributor agreement and pushed a second commit.  I don't know what the proper etiquette is here.  Feel free to squash it with my last one, or I can squash and re-write the branch history so it shows up as a single commit.

Stewori:
No worries. We usually squash it, so it should be fine however you provide the PR. @jeff5 Since this fix is really trivial I'll merge it right away. Hope that's fine :)

PR #343, opened 2024-07-18
export xml.parsers and xml.etree like Python 2.7

thetric:
Export the same modules as the upstream, see https://github.com/python/cpython/blob/v2.7.18/Lib/xml/__init__.py

### DIFF
--- a/Lib/xml/__init__.py
+++ b/Lib/xml/__init__.py
@@ -1,18 +1,23 @@
 """Core XML support for Jython.
 
-This package contains two sub-packages:
+This package contains four sub-packages:
 
 dom -- The W3C Document Object Model.  This supports DOM Level 1 +
        Namespaces.
+
+parsers -- Python wrappers for XML parsers (currently only supports Expat).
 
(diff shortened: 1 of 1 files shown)


Stewori:
This looks okay to me and should be a minimal, rather safe change (pointing this out since we are close to 2.7.4 release).
@jeff5 Is it okay to merge this now (there will be rc phase afterall)?

jeff5:
I don't know why these are suppressed (maybe just an old stdlib?).

It does seem minor, but we've be bitten by late change before. Exposing two new modules sounds quite significant. But we just mean that `from xml import *` makes them appear, don't we?

I know there are tests of `etree`because it always complains. https://github.com/jython/jython/actions/runs/9989854845/job/27666629754?pr=343#step:6:540

jeff5:
I think I understand this now as minimal impact. I also checked the test coverage, but we don't appear to lack anything major that is in `lib-python`, and I'm not going to mess with the differences there are.

Issue #344, opened 2024-07-19
Maintain ucnhash.dat


Stewori:
ucnhash.dat is required for the API in the [ucnhash module](https://apidia.net/java/Jython/2.7.3/?pck=org.python.modules&cls=.ucnhash), i.e. for looking up unicode characters by name. It should be updated for every new unicode release. Since Jython 2.7.4 release is approaching, it occurred to me that it has not been updated since 2017 for Unicode 9.0. Current release is Unicode 15.1.
I am preparing a PR for the update right now. The process is somewhat documented in `Misc/makeucnhashdat.py`, but I can write up more detailed instructions for the dev guide later.
One reason to open this issue is to have a proper associated issue number to reference in NEWS:
Updates to ucnhash.dat should be tracked somewhere and this can establish the convention to track them at least in NEWS with the canonical number of this issue, i.e. write a line
`    - [ GH-344 ] Updated ucnhash.dat to Unicode x.y.z`

Stewori:
Update to Unicode 15.1 in PR #345.

jeff5:
Does that close this?

If a change is minor enough (not saying this one is) I think opening a PR and citing that as the issue number in NEWS is ok. The numbers come from the same series.

It goes wrong where what seems minor to an inexperienced contributor actually turns out to require substantial discussion, which is then lost when we reject the PR to take a fresh approach. So it's right to err on the side of issue vs PR.

Stewori:
Yes, I suppose it can be closed. There is no rule against using closed issue numbers in NEWS. One use case of keeping it open would be to discuss whether we might want to switch to using [getCharFromName](https://apidia.net/mvn/com.ibm.icu/icu4j/74.2/?pck=com.ibm.icu.lang&cls=.UCharacter#getCharFromName-java.lang.String-) from [ICU4J](https://apidia.net/mvn/com.ibm.icu/icu4j). However, the discussion can be cut short since its not worth changing this for Jython 2.7 series and the discussion probably comes too early for Jython 3 series. So, no point in keeping this open now. Do we somewhere have a release guide? Like steps to do before a release, e.g. update dependencies etc? If so, there should be an entry/reminder with instructions to update ucn hash.

jeff5:
> Do we somewhere have a release guide?

[We do](https://jython-devguide.readthedocs.io/en/latest/release_jy.html), and I'm editing it right now, as I tend to with each release, and I have added a section for "Before you start".

[...]

Issue #342, opened 2024-07-18
Update dependencies


thetric:
As mentioned in https://github.com/jython/jython/pull/169#issuecomment-1133988745 and https://github.com/jython/jython/pull/157#issuecomment-1132604803 it is preferrable if team members update dependencies:

The following dependencies have later milestone versions:
 - [x] com.github.jnr:jffi [1.3.9 -> 1.3.13]
     http://github.com/jnr/jffi
 - [x] com.github.jnr:jnr-constants [0.10.3 -> 0.10.4]
     http://github.com/jnr/jnr-constants
 - [x] com.github.jnr:jnr-ffi [2.2.12 -> 2.2.16]
     http://github.com/jnr/jnr-ffi
 - [x] com.github.jnr:jnr-posix [3.1.15 -> 3.1.19]
 - [x] com.google.guava:guava [32.1.2-jre -> 33.2.1-jre]
     https://github.com/google/guava
 - [x] com.ibm.icu:icu4j [71.1 -> 75.1]
     https://icu.unicode.org/
 - [x] com.ibm.informix:jdbc [4.50.8 -> 4.50.11]
     https://www.ibm.com/informix
 - [x] com.oracle.database.jdbc:ojdbc8 [19.14.0.0 -> 23.4.0.24.05]
     https://www.oracle.com/database/technologies/maven-central-guide.html
 - [x] commons-io:commons-io [2.15.1 -> 2.16.1]
     https://commons.apache.org/proper/commons-io/
 - [ ] io.netty:netty-buffer [4.1.73.Final -> 4.1.111.Final]
     http://netty.io/
 - [ ] io.netty:netty-codec [4.1.73.Final -> 4.1.111.Final]
     http://netty.io/
 - [ ] io.netty:netty-common [4.1.73.Final -> 4.1.111.Final]
     http://netty.io/
 - [ ] io.netty:netty-handler [4.1.73.Final -> 4.1.111.Final]
     http://netty.io/
 - [ ] io.netty:netty-resolver [4.1.73.Final -> 4.1.111.Final]
     http://netty.io/
 - [ ] io.netty:netty-transport [4.1.73.Final -> 4.1.111.Final]
     http://netty.io/
 - [x] jline:jline [2.14.5 -> 2.14.6]
 - [x] org.antlr:antlr-runtime [3.5.2 -> 3.5.3]
     http://www.antlr.org
 - [x] org.apache.ant:ant [1.10.12 -> 1.10.14]
     https://ant.apache.org/
[...]

PR #345, opened 2024-07-19
Updated ucnhash.dat to Unicode 15.1.

Stewori:


Stewori:
Handles #344 for Unicode 15.1.

jeff5:
What's the relationship between this and ICU4J, also in line for update under #342? Do they need to update together? Or ICU4J before the other is generated?

Stewori:
I think that's independent. The ucnhash module does not use external API. It is a self-contained lookup based on a minimal perfect hash. It really just realizes the lookup of a unicode character by name based on data from UnicodeData.txt bundled with every official unicode release, e.g. https://www.unicode.org/Public/15.1.0/ucd/UnicodeData.txt If this update is forgotten, the only thing that happens is that the lookup does not recognize newer character names. Usually not so critical. I stumbled onto this, because I once worked out SymPy support and SymPy looked up some mathematical symbols by name and Jython's support was insufficient for SymPy (stood at unicode 3.something when there was 9.0 released). Had to figure out how to update the hash function that time. It might be that ICU features a similar API that could replace our hashing approach. So far I just did not want to touch the running system.

jeff5:
Git is giving me a by-line on this commit, which I didn't notice it was going to do. Didn't mean to steal your thunder.

Stewori:
NP. Don't care much about my Github reputation these days. If you want to make up for it, you can star the [APIdia repo](https://github.com/APIdia-net/apidia.net) :)

PR #355, opened 2024-08-31
Update urlparse.py

djallalzoldik:
Hi , team

i was able to convert urlparse.py from urllib that in python3 > https://raw.githubusercontent.com/python/cpython/3.9/Lib/urllib/parse.py

i was able to convert it to be compatible with pytho2
i already test it and it work fine

### DIFF
--- a/lib-python/2.7/base58.py
+++ b/lib-python/2.7/base58.py
@@ -1,0 +1,173 @@
+#cloned from https://github.com/keis/base58/blob/master/base58.py
+
+'''Base58 encoding 
+Implementations of Base58 and Base58Check endcodings that are compatible
+with the bitcoin network.
+'''
+
+# This module is based upon base58 snippets found scattered over many bitcoin
+# tools written in python. From what I gather the original source is from a
+# forum post by Gavin Andresen, so direct your praise to him.
+# This module adds shiny packaging and support for python3.
+
+from hashlib import sha256
+
+__version__ = '0.2.5'
+
+# 58 character alphabet used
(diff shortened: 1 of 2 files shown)


jeff5:
Thanks for offering this and for the work behind it.

We haven't previously backported code from Python 3, and I'm wary of starting now. What's the motivation? How much of the URL universe does the existing version not parse? If you have a particular need in your application, there are always ways to patch in an improvement as a local library.

Technicalities if we want to consider this:

- It's in the wrong place. `lib-python` is code we have copied from CPython for re-use and nobody changes it. What you would do is place the same module in the `~/Lib` tree, then the build would use that instead.
- As I understand it, urllib has grown significantly from Python 2 to Python 3. We would want to update the corresponding tests to Python 3. It may work fine for you, but we need to be able to test it every time we build. Again, these belong in the  `~/Lib` tree. This may be a difficult conversion because the test support has changed a lot.
- Your starting version seems to be before the changes that finally landed https://github.com/python/cpython/issues/102153 . I don't know if the current Python 2 library has the same faults, but it seems unwise to expand the range of protocols without fixing a well-publicised CVE.
- If I haven't put you off by now, there's a [contributor agreement](https://www.python.org/psf/contrib/contrib-form/) to sign.

jeff5:
Baseline would be that it passes the existing tests (unless there's a good reason why they're wrong).

lib-python/2.7/base58.py
jeff5, line 1: This does not appear to have anything to do with the PR.

djallalzoldik: I think all types of encoding are important. for the PR , specially in the new technologies or for the web3

PR #361, opened 2024-10-18
Remove synchronization of JyAttribute.getAttr to improve performance in multithreaded scenarios.

### LINKED ISSUE #360: JyAttribute.getAttr significantly slow down during multi-threading
I have a service deployed on AWS and use CodeGuru to analyze the performance. I have found when high concurrency >20TPS, `JyAttribute.getAttr` makes CPU in 95% of **BLOCKED** time. There was an issue in 2016, but I do not think it got resolved completely as there is still a Synchronized(ob) https://bugs.jython.org/issue2452.

This brought a huge latency impact due to synchronized. I am looking for a way to improve the performance during high-concurrency to avoid this kinds of lock. Thanks.

Stewori:
Fixes #360

### DIFF
--- a/src/org/python/core/JyAttribute.java
+++ b/src/org/python/core/JyAttribute.java
@@ -3,7 +3,6 @@ package org.python.core;
 import java.io.Serializable;
 
 /**
- * <p>
  * Manages a linked list of general purpose Object-attributes that
  * can be attached to arbitrary {@link org.python.core.PyObject}s.
  * This method replaces the formerly used method of maintaining weak
@@ -13,12 +12,10 @@ import java.io.Serializable;
  * to attach
  * {@link org.python.modules._weakref.GlobalRef}-objects in the
  * {@link org.python.modules._weakref.WeakrefModule}.
- * </p>
  * <p>
  * Attributes attached via the weak hash-map-method break, if the
  * {@code PyObject} is resurrected in its finalizer. The
  * {@code JyAttribute}-method is resurrection-safe.
- * </p>
  * <p>
  * To reduce memory footprint of {@code PyObject}s, the fields for
  * {@link org.python.core.finalization.FinalizeTrigger}s and
@@ -26,7 +23,6 @@ import java.io.Serializable;
  * on top so there is no speed-regression,
  * {@link org.python.core.finalization.FinalizeTrigger} on bottom,
  * as it is usually never accessed.
- * </p>
  */
 public abstract class JyAttribute implements Serializable {
     /* ordered by priority; indices >= 0 indicate transient attributes.
@@ -48,7 +44,7 @@ public abstract class JyAttribute implements Serializable {
     public static final byte WEAK_REF_ATTR = 0; //first transient
 
     /**
-     * Reserved for use by <a href="http://www.jyni.org" target="_blank">JyNI</a>.
+     * Reserved for use by <a href="https://jyni.12hp.de" target="_blank">JyNI</a>.
      */
     public static final byte JYNI_HANDLE_ATTR = 1;
 
@@ -87,15 +83,15 @@ public abstract class JyAttribute implements Serializable {
     public static final byte GC_DELAYED_FINALIZE_CRITICAL_MARK_ATTR = 6;
 
     public static final byte FINALIZE_TRIGGER_ATTR = Byte.MAX_VALUE;
-    private static byte nonBuiltinAttrTypeOffset = Byte.MIN_VALUE+1;
-    private static byte nonBuiltinTransientAttrTypeOffset = 7;
+    private static volatile byte nonBuiltinAttrTypeOffset = Byte.MIN_VALUE+1;
+    private static volatile byte nonBuiltinTransientAttrTypeOffset = 7;
 
     /**
      * Reserves and returns a new non-transient attr type for custom use.
      *
      * @return a non-transient attr type for custom use
      */
-    public static byte reserveCustomAttrType() {
+    public static synchronized byte reserveCustomAttrType() {
         if (nonBuiltinAttrTypeOffset == 0) {
             throw new IllegalStateException("No more attr types available.");
         }
@@ -107,18 +103,18 @@ public abstract class JyAttribute implements Serializable {
      *
      * @return a transient attr type for custom use
      */
-    public static byte reserveTransientCustomAttrType() {
+    public static synchronized byte reserveTransientCustomAttrType() {
         if (nonBuiltinTransientAttrTypeOffset == Byte.MAX_VALUE) {
             throw new IllegalStateException("No more transient attr types available.");
         }
         return nonBuiltinTransientAttrTypeOffset++;
     }
 
-    byte attr_type;
+    final byte attr_type;
 
     static class AttributeLink extends JyAttribute {
-        JyAttribute next;
-        Object value;
+        volatile JyAttribute next;
+        volatile Object value;
 
         protected AttributeLink(byte attr_type, Object value) {
             super(attr_type);
@@ -147,8 +143,8 @@ public abstract class JyAttribute implements Serializable {
     }
 
     static class TransientAttributeLink extends JyAttribute {
-        transient JyAttribute next;
-        transient Object value;
+        volatile transient JyAttribute next;
+        volatile transient Object value;
 
         protected TransientAttributeLink(byte attr_type, Object value) {
             super(attr_type);
@@ -197,18 +193,51 @@ public abstract class JyAttribute implements Serializable {
      * Retrieves the attribute of the given type from the given
      * {@link org.python.core.PyObject}.
      * If no attribute of the given type is attached, null is returned.
+     * 
+     * This class is generally thread safe, but this method is not synchronized with
+     * concurrent write access to avoid performance issues for multithreaded scenarios.
+     * The trade-off for that choice is that in case of concurrent write access, it is
+     * unspecified whether this method returns according to the old state or the new state.
+     * This method is thread safe in the sense that a concurrent write access never exposes
+     * an invalid intermediate state to this read access.
      */
     public static Object getAttr(PyObject ob, byte attr_type) {
-        synchronized (ob) {
-            if (ob.attributes instanceof JyAttribute) {
-                JyAttribute att = (JyAttribute) ob.attributes;
-                while (att != null && att.attr_type < attr_type) {
-                    att = att.getNext();
-                }
-                return att != null && att.attr_type == attr_type ? att.getValue() : null;
+        /*
+        Here used to be a synchronized(ob) block encapsulating the whole method code.
+        It has been removed due to performance issues reported in
+        github.com/jython/jython/issues/360.
+        By making ob.attributes in PyObject and next and value in AttributeLink and
+        TransientAttributeLink volatile, the read access in this method does not
+        fatally interfere with a concurrent write operation.
+        (Write operations continue to be synchronized against each other.)
+        https://docs.oracle.com/javase/specs/jls/se17/html/jls-17.html#jls-17.4.4 states:
+        "A write to a volatile variable v (§8.3.1.4) synchronizes-with all subsequent reads
+         of v by any thread"
+        This induces an x happens-before y relation hb(x, y) according to
+        https://docs.oracle.com/javase/specs/jls/se17/html/jls-17.html#jls-17.4.5:
+        "If an action x synchronizes-with a following action y, then we also have hb(x, y)."
+        Also see https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html,
+        which confirms the anticipated behavior of volatile fields for Java 5 and onwards.
+        Write accesses in the linked attribute list are designed to update the list tail
+        such that a concurrent read access would iterate either through the old state or
+        the new state. No invalid intermediate state is exposed to the read access.
+        Due to volatile semantics, this assumption cannot be broken by code reordering
+        issues like pointed out in the linked DoubleCheckedLocking document.
+        */
+        Object oatt = ob.attributes; // atomic one-time access to ob.attributes, which is volatile
+        if (oatt instanceof JyAttribute) {
+            JyAttribute att = (JyAttribute) oatt;
+            while (att != null && att.attr_type < attr_type) {
+                att = att.getNext(); // atomic access to next, which is volatile
+                // Perhaps att is affected by a concurrent write access, perhaps not.
+                // Since the next-field is volatile, att is always a properly initialized
+                // attribute sequence head at this point.
+                // The overall iteration is therefore either according to the old state
+                // or according to the new state.
             }
-            return attr_type == JAVA_PROXY_ATTR ? ob.attributes : null;
+            return att != null && att.attr_type == attr_type ? att.getValue() : null;
         }
+        return attr_type == JAVA_PROXY_ATTR ? oatt : null;
     }
 
     /**
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # Jython: Python for the Java Platform
 [![Maven Central](https://maven-badges.herokuapp.com/maven-central/org.python/jython-standalone/badge.svg)](https://search.maven.org/artifact/org.python/jython-standalone/)
 [![Javadocs](https://www.javadoc.io/badge/org.python/jython-standalone.svg)](https://www.javadoc.io/doc/org.python/jython-standalone)
+[![APIdia](https://apidia.net/java/Jython/badge.svg)](https://apidia.net/java/Jython)
 
 This is the development repository of Jython,
 the implementation of Python in Java.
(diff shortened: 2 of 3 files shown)


src/org/python/core/JyAttribute.java
```diff
@@ -197,18 +193,51 @@ public static boolean hasAttr(PyObject ob, byte attr_type) {
...
+        (Write operations continue to be synchronized against each other.)
+        https://docs.oracle.com/javase/specs/jls/se17/html/jls-17.html#jls-17.4.4 states:
+        "A write to a volatile variable v (§8.3.1.4) synchronizes-with all subsequent reads
+         of v by any thread"
+        This induces an x happens-before y relation hb(x, y) according to
+        https://docs.oracle.com/javase/specs/jls/se17/html/jls-17.html#jls-17.4.5:
+        "If an action x synchronizes-with a following action y, then we also have hb(x, y)."
+        Also see https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html,
+        which confirms the anticipated behavior of volatile fields for Java 5 and onwards.
+        Write accesses in the linked attribute list are designed to update the list tail
```
jeff5, line 221: Isn't this updating the list head? https://github.com/jython/jython/blob/4c271d88e6a85847b5f2d378cdbc7b3553aab2ae/src/org/python/core/JyAttribute.java#L296-L300
I don't follow all the logic yet, but I think you are keeping them in order of attribute type (by insertion sort).

Looking around a bit, I think it is essential to your argument that the reader effectively sees the old list, even while (one) writer is creating a modified one. You can also modify a node in the middle, and I can't decide whether `volatile` then saves you.

src/org/python/core/JyAttribute.java
```diff
@@ -197,18 +193,51 @@ public static boolean hasAttr(PyObject ob, byte attr_type) {
...
+        such that a concurrent read access would iterate either through the old state or
+        the new state. No invalid intermediate state is exposed to the read access.
```
jeff5, line 223: It will be important that the list node is fully initialised and stable before it can be referenced by the unsynchronised reader in code like this: https://github.com/jython/jython/blob/4c271d88e6a85847b5f2d378cdbc7b3553aab2ae/src/org/python/core/JyAttribute.java#L296-L300
I see that the lines of code here are sequenced so that the publication of the reference to the new node comes last. Now we have to be sure the compiler won't rearrange assignment before construction. ISTR either the JLS or Pugh explicitly mentions this possibility. We have guarantees about writes to an individual variable, but what about actions nearby in the programme order of the thread? I think the generous application of the `volatile` keyword may have achieved that, but as I've said, I find this stuff difficult to reason about securely.

src/org/python/core/JyAttribute.java
```diff
@@ -197,18 +193,51 @@ public static boolean hasAttr(PyObject ob, byte attr_type) {
...
+        Due to volatile semantics, this assumption cannot be broken by code reordering
+        issues like pointed out in the linked DoubleCheckedLocking document.
+        */
+        Object oatt = ob.attributes; // atomic one-time access to ob.attributes, which is volatile
+        if (oatt instanceof JyAttribute) {
+            JyAttribute att = (JyAttribute) oatt;
+            while (att != null && att.attr_type < attr_type) {
+                att = att.getNext(); // atomic access to next, which is volatile
+                // Perhaps att is affected by a concurrent write access, perhaps not.
```
jeff5, line 232: Block comment better.

jeff5 (changes requested):
It all hangs on the correctness of the argument. I find these things hard to convince myself of from just the theory.

An example of a test that arranges intense thread competition is [this one I wrote to hammer the (prospective) type system](https://github.com/jeff5/very-slow-jython/blob/5a0c4b48759999d045ef2af1665cd26cc1d254b6/rt4core/src/kernelTest/java/uk/co/farowl/vsj4/runtime/BootstrapTest.java#L257-L277). There was a risk, if I hadn't got it right, of deadlock or publication of half-finished objects.

The time measurements are made to prove the order of events rather than to measure performance. Synchronisation is explicit, so I'm not too worried about code re-ordering in my case.

The claim that the reader sees a coherent state as last written would be easier to accept if `setAttr` were to make a new list, in which everything could be constructed `final`, and only the head pointer would need to be `volatile` (and the `nonBuiltin*` counters). Assuming a `setAttr` is much less common than a `getAttr`, and lists are mostly short, the copy would not be expensive. You are best placed to say whether that's true.

src/org/python/core/JyAttribute.java
Stewori: As I understand the volatile semantics, having `ob.attributes` and  (`Transient`)`AttributeLink.next` volatile should indeed prevent reordering of `newAtt.setNext(att);` and `ob.attributes = newAtt;`. Do you have an idea how we can confirm that? Perhaps by disassembling the bytecode of `JyAttribute.class`. That would be tedious and I suspect the JIT may apply subsequent optimization, including reordering where permitted. Perhaps an isolated test of `JyAttribute` concurrent access? Like 100 Threads read and write attributes permanently for 5 minutes for a dummy `PyObject`. This really stresses the design and the test would observe the results to confirm the anticipated behavior. Still tedious to implement. Read access would have to confirm that concurrent write access would never make an unrelated attribute (with higher index than the written one) temporarily invisible. Provoking an exception of any kind would of course be a fast-fail for the whole test.

I'm skeptical about changing the implementation more broadly (to create an entirely new list every time). If the volatile semantics does not work as expected and promised, that approach would be similarly broken. Proper initialization of the whole new list might be reordered with assignment to `ob.attribute`. A read access in the wrong moment could pick the uninitialized thing. So either it works as specified and it would be fine in any case, or it would be broken in any case. The new-list-every-time-approach would btw also require `next` to be volatile to prevent reordering of list initialization. Then, if we would make a new thing every time we could also make it an array. But also then, the array would have to be volatile to ensure proper initialization. However I'm not even aware what volatile would *theoretically* mean for an array (every item be like a volatile field I suppose).

All in all, the question of trust in the specs applies to each approach and we would likely need a proper stress test in any case. So I would go with the already implemented solution, wich is of all alternatives the one with the fewest changes to the status quo. Also, I do not see enough benefit to justify rewriting it yet again.

src/org/python/core/JyAttribute.java
Stewori: According to the specs, I think it would (volatile save me, that is). It should prevent reordering `newAtt.setNext` (because all it does is assign the arg to the volatile field `next`) with `ob.attributes` assignment. If a node in the middle is inserted, the reader might either be still before that node or have already passed it. Therefore, it is unspecified whether according to the old or new state is returned. The approach should, however, ensure that in no case unrelated attributes are temporarily unreachable, because the tail is already attached to the new middle node before it is inserted with `volatile` preventing reordering.

Stewori:
Before we think about writing a stress test I suggest to await response of @newprogrammer101 . If the PR does not solve the issue or blows something up, we would need to rethink the solution anyway.

src/org/python/core/JyAttribute.java
Stewori: Wish shall be granted :)

src/org/python/core/JyAttribute.java
jeff5: The passage (JLS 17.4.5) that nags me a bit is:

>  if two actions share a happens-before relationship, they do not necessarily have to appear to have happened in that order to any code with which they do not share a happens-before relationship. Writes in one thread that are in a data race with reads in another thread may, for example, appear to occur out of order to those reads.

but I think you are right that because `newAtt.next = att` and `ob.attributes = newAtt` are both (inter-thread) actions thanks to `volatile`, the order of writes in the writer and the reader must be the same. Similarly, the order of reads in the reader must be the same to both threads. So if a read of `ob.attributes` yields `newAtt`, rather than a previous value, then `newAtt.next` must by then be `att` as desired.

What about `newAtt.attr_type`: might it still have its default value `0` because `newAtt.attr_type = attr_type` was reordered after the other writes? Accoding to Pugh, we are not guaranteed that initialisation of `newAtt` has completed.

src/org/python/core/JyAttribute.java
jeff5: Good point about the array. And given that initialisation might skip round the write to `ob.attributes` I'm not convinced it helps. The idea was to reduce the number of volatile fields so we had only one to reason about instead of 3n+1 things.

That's the sort of test I meant, but it might not have to run so long if you could guarantee the threads compete. Inspecting the output of a particular compiler is not good enough. Our correctness has to be based on good theory and a reasonably harsh practical exam.

src/org/python/core/JyAttribute.java
Stewori: Re `attr_type`, I somehow thought that making it `final` would ensure high priority on initialization. Like that `final` would imply `volatile` semantics. If there is a point in making a field `final volatile` (and if that's even possible), then sure, we can add the volatile-bit to the `attr_type` field.

Stewori: From [JLS 17.5](https://docs.oracle.com/javase/specs/jls/se17/html/jls-17.html#jls-17.5):
> An object is considered to be completely initialized when its constructor finishes. A thread that can only see a reference to an object after that object has been completely initialized is guaranteed to see the correctly initialized values for that object's final fields.

Due to `volatile` preventing reordering, the constructor of `AttributeLink` should have finished before its assignment to `next` and therefore, the final field should be properly initialized.

src/org/python/core/JyAttribute.java
```diff
@@ -87,15 +83,15 @@ public abstract class JyAttribute implements Serializable {
...
-    private static byte nonBuiltinTransientAttrTypeOffset = 7;
+    private static volatile byte nonBuiltinAttrTypeOffset = Byte.MIN_VALUE+1;
+    private static volatile byte nonBuiltinTransientAttrTypeOffset = 7;
 
     /**
      * Reserves and returns a new non-transient attr type for custom use.
      *
      * @return a non-transient attr type for custom use
      */
-    public static byte reserveCustomAttrType() {
```
fayazkhan121, line 98: Observation: The method reserveCustomAttrType has been changed to a synchronised method. This is a good practice for thread safety, but it introduces a potential bottleneck under heavy load.

Recommendation: Assess if finer-grained synchronisation (e.g., using atomic variables) could replace the synchronisation here to improve concurrency without sacrificing thread safety.

src/org/python/core/JyAttribute.java
```diff
@@ -87,15 +83,15 @@ public abstract class JyAttribute implements Serializable {
...
+    public static synchronized byte reserveCustomAttrType() {
         if (nonBuiltinAttrTypeOffset == 0) {
```
fayazkhan121, line 95: Magic numbers are being used in conditional checks.
Recommended : Define these values as constants to improve code readability and maintainability:

fayazkhan121 (commented):
`if (nonBuiltinAttrTypeOffset == 0) {
    throw new IllegalStateException("No more attr types available.");
}
`
Magic numbers are being used in conditional checks.
Recommendation: Define these values as constants to improve code readability and maintainability

public static byte reserveCustomAttrType() {
    public static synchronized byte reserveCustomAttrType() {
The method reserveCustomAttrType has been changed to a synchronised method. This is a good practice for thread safety, but it introduces a potential bottleneck under heavy load.
Recommendation: Assess if finer-grained synchronisation (e.g., using atomic variables) could replace the synchronisation here to improve concurrency without sacrificing thread safety.

src/org/python/core/JyAttribute.java
jeff5: AI? Infrequently called, and relieves bottlenecks elsewhere. (Point of PR.)

src/org/python/core/JyAttribute.java
jeff5: May be valid, but not in scope of change.

newprogrammer101:
> Before we think about writing a stress test I suggest to await response of @newprogrammer101 . If the PR does not solve the issue or blows something up, we would need to rethink the solution anyway.

Thanks @Stewori for the quick change in this CR. This commit is to resolve concurrency issue on Jython package. But currently when I created multi JythonExecutor instance for testing purpose, I found CPU was at BLOCK state on `PrintStream.flush` as mentioned in https://github.com/jython/jython/issues/360#issuecomment-2550203822. I would suggest re-evaluate not only synchronize key word on Jython package, but also on its dependencies. @jeff5

Stewori:
I finally took the time to stress test the implementation and it works like expected. Generally, the implementation performs nicely. Trouble is, it also performs nicely *without* using `volatile`. So, the challenge was to get the stress test to actually provoke a failure if `volatile` is removed, and then to observe that the failure goes away if `volatile` is added back. My test now achieves exactly this. However, I don't think it is suitable to be included into the normal test suite, but I will share it here, so whoever is interested (@jeff5 ) can try it.

With that, I'm now even more confident in this PR and think it should be included into the 2.7.5 release.

Stewori:
[test_JyAttribute.zip](https://github.com/user-attachments/files/28318015/test_JyAttribute.zip)

Okay, `java DictStressTest` runs the test. It is very memory hungry, because for sufficient stress I had to remove all waiting entirely (sorry the code might not be in perfect shape and I probably mixed tabs and spaces...). On my machine, it can consume up to 4GB of ram. In theory, the test should run only one second, but in fact, it takes much longer (the shutdown command is delayed due to all the stress).
At the end, you should see an output like this:
```
...
rote(8): 511
wrote(1): 251
wrote(2): 918
Number of items of key 1: 9635557
Number of violations of key 1: 0
Number of items of key 2: 8467645
Number of violations of key 2: 0
Number of items of key 3: 8461724
Number of violations of key 3: 0
Number of items of key 4: 8091572
Number of violations of key 4: 0
Number of items of key 5: 6257706
Number of violations of key 5: 0
Number of items of key 6: 5352900
Number of violations of key 6: 0
Number of items of key 7: 4274878
Number of violations of key 7: 0
Number of items of key 8: 4562908
Number of violations of key 8: 0
```
And no violations means that the test passed. The number of items for each key indicates how many accesses to the dict have been recorded and evaluated for consistency.
Now, in `JyAttribute.java`, out-comment these lines:
```
        volatile // <- comment out this line to see failures
        transient JyAttribute next;
        volatile // <- comment out this line to see failures
        transient Object value;
```
Then recompile (`javac *.java`) and repeat the test. *Often*, it will be just fine, but try it ten or more times and occasionally you should see an output like this:
```
...
wrote(5): 409
wrote(6): 908
wrote(8): 396
wrote(7): 301
wrote(4): 10
Number of items of key 1: 9389605
Number of violations of key 1: 9
Number of items of key 2: 7791174
Number of violations of key 2: 13
Number of items of key 3: 6184843
Number of violations of key 3: 1
Number of items of key 4: 6899296
Number of violations of key 4: 7
Number of items of key 5: 6769969
Number of violations of key 5: 7
Number of items of key 6: 6066541
Number of violations of key 6: 14
Number of items of key 7: 4353699
Number of violations of key 7: 0
Number of items of key 8: 4135389
Number of violations of key 8: 0
```
which confirms that the test is sharp. In all tries so far (perhaps ~50) I have not encountered a single failure with `volatile` in place.

jeff5:
Ok, I believe.

Since we first considered this I have been back to the textbooks on this difficult subject. I see how `volatile` probably does the trick, but I think the test was worth it, thanks. Although in another context ... I can't seem to get test failures from deliberately incorrect synchronisation.

Do you think this should be a test in `ant javatests`? It would then be run routinely on a variety of architectures.

Stewori:
> Although in another context ... I can't seem to get test failures from deliberately incorrect synchronisation.

Please, share more details. What other context?

> Do you think this should be a test in ant javatests? It would then be run routinely on a variety of architectures.

No, it is too resource hungry and too non-telling. I think the stress level for a meaningful test is machine dependent and one would have to assert in non-volatile version that the test triggers. To know that the stress is meaningful. I don't see how this check could be done on a regular basis in normal tests. In fact like one in ten times it crashes my JVM with out-of-memory error. Maybe it would work better if the number of calls is limited rather than the time. Because the code that terminates it after the specified time lags behind due to stress. OTOH, if this is relieved, we might stop seeing the failures again. It would be tricky to tune the test to reliably trigger a violation on every machine. Especially without wasting too many resources. Perhaps it could be added as optional test. We should definitely make it available for off-stream testing and investigations. However, maybe it's enough to just link to my previous post from the source code (perhaps even from a doc comment) in `JyAttribute.java`. And maybe add a hint to the devguide or userguide.

src/org/python/core/JyAttribute.java
jeff5: I'm more comfortable withthe theory now. And given the practical test has been passed I think we're ok.

jeff5:
> > Although in another context ... I can't seem to get test failures from deliberately incorrect synchronisation.
> 
> Please, share more details. What other context?

I was referring to my unsuccessful attempts to produce a test-failing race on the type system in my Very slow Jython project. It's not directly relevant, except to say that it is a difficult area (for me at least) in which to construct a test guaranteed to reveal a problem.

PR #393, opened 2025-08-04
Some refactorings

thetric:
These are mostly automated refactorings, best review each commit

### DIFF
--- a/src/com/xhaus/modjy/ModjyJServlet.java
+++ b/src/com/xhaus/modjy/ModjyJServlet.java
@@ -124,7 +124,7 @@ public class ModjyJServlet extends HttpServlet {
             modjyServlet = (HttpServlet)temp;
             modjyServlet.init(this);
         } catch (PyException pyx) {
-            throw new ServletException("Exception creating modjy servlet: " + pyx.toString(), pyx);
+            throw new ServletException("Exception creating modjy servlet: " + pyx, pyx);
         }
     }
 
@@ -250,7 +250,7 @@ public class ModjyJServlet extends HttpServlet {
                 systemState.path.append(new PyString(archiveRealpath));
             }
         } catch (IOException iox) {
-            System.err.println("IOException: " + iox.toString());
+            System.err.println("IOException: " + iox);
         }
     }
 }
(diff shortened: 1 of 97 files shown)


Stewori:
I have mixed feelings about this. It fixes things that are not broken. It is a very big change set and we want to keep changes in Jython 2 minimal since it's kind of in maintenance mode. OTOH this clearly modernizes the code and most changes are justified. Some might even benefit maintainability. So, this might be mainly a matter of taste. Perhaps an unnecessary risk, but then we have tests to secure such changes. Then again, I wouldn't want someone to spend too much time (or time at all) on reviewing this PR since it doesn't really fix anything.

A more substantial concern is that it probably bumps the minimally required Java version heavily. E.g. I don't think try-with-resources or replacing if with switch in that way would be supported on Java 8 (or 9). There are surely good reasons to increase the required Java version, but cosmetic changes are not among these.

So, I'd say if it's only about the subset that leaves the currently required Java version unchanged, we might consider it.

thetric:
Every change should be compatible with Java 8. This can be verified by either compling with Java 8 or using the `--release 8` javac flag in Java 9 or higher.

Yeah, I might have gone a bit too far 😅 I am open to drop some unnecessary commits

Stewori:
Looks like `test_concat` and `test_random` are failing unexpectedly. Ideas?

thetric:
I guess the `test_random` is failing because of the switch vom `java.util.Random` to `java.security.SecureRandom`?

thetric:
The concat test seems to fail due to a buggy refactoring

Stewori:
> I guess the `test_random` is failing because of the switch vom `java.util.Random` to `java.security.SecureRandom`?

Please make the switch to `SecureRandom` a separate PR:

- the test failure indicates it might be more than a refactoring
- it relieves this PR (wich is already inconveniently big) by one test failure
- the decision about that switch and handling the test failure would be disentangled from this PR

thetric:
I removed some of the more stylistic refactorings and the switch to SecureRandom. Now the diff of the changeset is about half as big

thetric:
Ah, the test `test_fileutil_wrap_outputstream_default_textmode` fails because setting the system property `line.separator` in https://github.com/jython/jython/blob/3760bfb962803cff718fc7cd5ba658c9fd467fec/Lib/test/test_java_integration.py#L169 now has no effect

Stewori:
Please move refactorings that cause test failures into separate PRs. I acknowledge that in some cases the tests might require adjustment, but that requires a more careful case by case analysis, better suited in individual PRs.

thetric:
I will split up this large PR and reduce the overall changes to include only the relevant changes

PR #403, opened 2025-11-27
Increase the dependabot interval to 'weekly'

ohumbel:
Daily checks are not needed any more - save some resources.


ohumbel:
Sorry, should not have been submitted to upstream

PR #415, opened 2026-03-04
Add @Documented to all annotations in org.python.expose.

### LINKED ISSUE #388: @ExposedType annotation should be @Documented
[`@ExposedType`](https://apidia.net/java/Jython/2.7.4/?pck=org.python.core&cls=org.python.expose.ExposedType) should be [`@Documented`](https://apidia.net/java/OpenJDK/24/?pck=java.lang.annotation&cls=.Documented), because it's relevant for users to know and shouldn't require reading the source code. Same for `@ExposedMethod` etc. I self fell for this in https://github.com/jython/jython/issues/387#issuecomment-2940495624 where I suggested the wrong method even after looking into API docs. I might have noticed it easier if `@ExposedType` and `@ExposedMethod` were not hidden by default, which is a consequence of being considered not `@Documented`. (In the APIdia docs one can force-show annotations which are not `@Documented` via a filter in the top-right menu. Conventional Javadoc excludes them completely.)

Stewori:
Closes #388


Stewori:
This should be a safe one. (heard that before^^, little did they know) 
However, tests support this opinion. I'll leave it here for others to take a look for typos @jeff5 @ohumbel 
Will merge in a couple of days if nobody speaks up.

ohumbel:
I checked the changes very carefully and could not spot any typos...
Looks safe to me.

jeff5:
Yes to `@Documented` in general, but I see you've tidied up some Javadoc, so I'll read carefully.

src/org/python/expose/ExposedNew.java
```diff
@@ -4,27 +4,26 @@
...
- *<p>
  * In the first form, the method must be static and take the arguments
- * <code>PyNewWrapper new_, boolean init, PyType
- subtype, PyObject[] args, String[] keywords</code>. In this case, the method has full
- * responsibility for creating and initting the object and will be invoked for every subtype of this
- * exposed type. Essentially it's for object instantiation that must be called for every instance of
- * that object. See {@link PyInteger#int_new} for an example of this type of ExposedNew.
- *<p>
+ * {@code PyNewWrapper new_, boolean init, PyType subtype, PyObject[] args, String[] keywords}.
+ * In this case, the method has full responsibility for creating and initiating the object and
```
jeff5, line 15:
```suggestion
 * In this case, the method has full responsibility for creating and initialising the object and
```

jeff5: Except of course now we run into US vs UK English. (I won't object to either.)

src/org/python/expose/ExposedType.java
```diff
@@ -22,18 +24,18 @@
     String name() default "";
 
     /**
-     * @return the base type of this type. Must be another class anotated with ExposedType. If
-     *         unspecified, the base is set to object, or PyObject.class.
+     * @return the base type of this type. Must be another class annotated with {@code ExposedType}.
+     *         If unspecified, the base is set to {@code object}, or {@code PyObject.class}.
```
jeff5, line 28: If the `@return` has to be qualified I would probably give the full version in the Javadoc body, and a minimal one in the `@return`.

jeff5, line 28:
```suggestion
    * Return the base type of this type. This must be another class annotated with {@code ExposedType}.
     *         If not specified (or {@code Object.class} is given), the base is set to Python {@code object}, that is {@code PyObject.class}.
     * @return the base type of this type.
```

src/org/python/expose/ExposedNew.java
Stewori: I suppose, AE is more in line with PSF standard, so "initializing" it will be. Also, my dict says it's somewhat acceptable in BE unlike the other way round.

src/org/python/expose/ExposedType.java
Stewori: Yes, that's the typical way. My intention was to keep changes minimal and to bring everything into consistent form. Since some class used plainly `@return ...` without prior main-body doc, while other classes used `Returns ...` I decided to use the `@return` way consistently. If no body html is present, but *only* `@return ...` block, APIdia employs `@return ...` content also as main body prefixed with `Returns: ` and as `@return ...` block tag. So this leads to a proper doc. I used to think Javadoc would do it similarly. However, in conventional Javadoc, this became possible just with Java 16 which introduced the inline `{@return ...}` tag:

https://bugs.openjdk.org/browse/JDK-8075778?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel

So, if we would build with e.g. Java 17 or later, The solution would be using inline `{@return ...}`.

Stewori: Yes, this makes sense for the long description of `base()`. However the principle does not fit naturally to the other methods which have very short doc.

src/org/python/expose/ExposedType.java
```diff
@@ -22,18 +24,18 @@
...
 
     /**
-     * @return Whether this type allows subclassing.
+     * @return whether this type allows subclassing.
      */
     boolean isBaseType() default true;
 
     /**
-     * Returns the __doc__ String for this type.
+     * @return the {@code __doc__} string for this type.
```
Stewori: @jeff5 You see, here I changed it just for consistency. The previous method `isBaseType()` uses plainly `@return` in the first place. If we cannot use `{@return ...}`, maybe it would be best to make them all consistently `Returns ...`. Then, at least it's always readable in the summary, also in older Javadoc.

Stewori:
Okay, to move on and - I think doc-related concerns have been addressed, I'll merge this one now.

PR #419, opened 2026-03-31
Fix TypeError when using subprocess shell=True under strict Java Secu…

### LINKED ISSUE #418: TypeError: unsupported operand type(s) for +: 'NoneType' and 'list' in subprocess.Popen when using shell=True under strict Java SecurityManager
### Description

   When running Jython within a sandboxed JVM environment that utilizes a strict `java.lang.SecurityManager` (specifically, one that denies `java.io.FilePermission` for reading system directories like `/bin/`),
   executing commands via the `subprocess` module with `shell=True` (or using wrappers like `os.popen` or `os.system`) crashes with a cryptic Python `TypeError` rather than throwing a descriptive `OSError` or
   Java `SecurityException`.

   ### Environment
   *   **Jython Version:** 2.5.x, 2.7.x (including 2.7.3 and current development branch)
   *   **OS:** Linux/Unix (where `_shell_command` resolution relies on `/bin/sh`)
   *   **JVM State:** Running with a `SecurityManager` that denies read access to standard system executable paths.

   ### Steps to Reproduce
   Run a Jython script that attempts to use `os.popen()` or `subprocess.Popen(..., shell=True)` inside a JVM where the `SecurityManager` denies file read access to `/bin/sh`.

   **Minimal Jython Script:**
```python
  import java.lang.SecurityManager as SecurityManager
  import java.lang.System as System
  import java.security.AccessControlException as AccessControlException

  class MySecurityManager(SecurityManager):
      def checkRead(self, file, context=None):
          if file == "/bin/sh" or file == "cmd.exe" or file.endswith("sh"):
              raise AccessControlException("Access denied: " + file)
      def checkPermission(self, perm, context=None):
          pass

  System.setSecurityManager(MySecurityManager())
```

```python  
import subprocess
# This will crash with a TypeError
subprocess.Popen(["whoami"], shell=True)
```

```   
**Stack Trace:**
  test_sm.py:16: RuntimeWarning: Unable to determine _shell_command for underlying os: posix
    import subprocess
    File "test_sm.py", line 17, in <module>
      subprocess.Popen(["whoami"], shell=True)
    File ".../Lib/subprocess$py.class", line 892, in __init__
    File ".../Lib/subprocess$py.class", line 1352, in _execute_child
  TypeError: unsupported operand type(s) for +: 'NoneType' and 'list'
```

   ### Root Cause Analysis
   The bug resides in `Lib/subprocess.py` and stems from how Jython caches the default system shell executable during module initialization.

   1.  **Initialization (`_setup_platform`)**: When the `subprocess` module is loaded, it attempts to determine the system shell by iterating through known shell paths (e.g., `/bin/sh`). It validates these paths
   using `os.path.exists(executable)`.
   2.  **SecurityManager Intervention**: Because the JVM is sandboxed, `os.path.exists("/bin/sh")` triggers the `SecurityManager`, which throws an `AccessControlException`.
   3.  **Silent Failure**: Jython's `os.path.exists()` implementation catches this Java security exception and simply returns `False`.
   4.  **Invalid State**: Because `os.path.exists()` returns `False` for all candidate shells, the global variable `_shell_command` is never assigned a valid list (like `["/bin/sh", "-c"]`) and remains
   initialized to `None`.
   5.  **The Crash (`_execute_child` and `system`)**: Later, when the user invokes `subprocess.Popen(..., shell=True)` or `os.system()`, the code attempts to prepend the shell command to the user's arguments:
   ```python
      if shell:
          args = _shell_command + args # <--- CRASH HERE
   ```

       Because `_shell_command` is `None` and `args` is a list, Python throws the `TypeError`.

   ### Expected Behavior
   Instead of a cryptic `TypeError` about concatenating `NoneType` and `list`, Jython should gracefully handle the scenario where a shell cannot be located (whether due to actual absence or `SecurityManager`
   restrictions).

akspi:
This change addresses the issue captured here: https://github.com/jython/jython/issues/418

Fixes jython/jython#418

### DIFF
--- a/Lib/test/test_subprocess.py
+++ b/Lib/test/test_subprocess.py
@@ -724,10 +724,46 @@ class ProcessTestCase(unittest.TestCase):
                                "-c", 1])
 
 
+class SecurityManagerTestCase(unittest.TestCase):
+    def test_subprocess_shell_with_security_manager(self):
+        if not jython:
+            return
+
+        import java.lang.SecurityManager as SecurityManager
+        import java.lang.System as System
+        import java.security.AccessControlException as AccessControlException
+
+        class MySecurityManager(SecurityManager):
+            def checkRead(self, file, context=None):
+                if file == "/bin/sh" or file == "cmd.exe" or file.endswith("sh"):
+                    raise AccessControlException("Access denied: " + file)
+            def checkPermission(self, perm, context=None):
+                pass
+
+        original_sm = System.getSecurityManager()
+        System.setSecurityManager(MySecurityManager())
+        try:
+            # We must force re-initialization of subprocess _shell_command
+            # because it was already initialized when the module was loaded.
+            saved_shell_command = subprocess._shell_command
+            subprocess._shell_command = None
+            subprocess._setup_platform()
+            
+            try:
+                subprocess.Popen(["whoami"], shell=True)
+                self.fail("Expected OSError")
+            except OSError:
+                pass
+            except TypeError as e:
+                self.fail("Caught TypeError instead of OSError: " + str(e))
+        finally:
+            System.setSecurityManager(original_sm)
+            subprocess._shell_command = saved_shell_command
+
 def test_main():
     # Spawning many new jython processes takes a long time
     test_support.requires('subprocess')
-    test_support.run_unittest(ProcessTestCase)
+    test_support.run_unittest(ProcessTestCase, SecurityManagerTestCase)
     if hasattr(test_support, "reap_children"):
         test_support.reap_children()
 
--- a/Lib/subprocess.py
+++ b/Lib/subprocess.py
@@ -1349,6 +1349,8 @@ class Popen(object):
             args = _escape_args(args)
 
             if shell:
+                if _shell_command is None:
+                    raise OSError("Unable to locate or access a valid system shell to execute shell=True commands.")
                 args = _shell_command + args
 
             if executable is not None:
@@ -1936,6 +1938,8 @@ def _os_system(command):
     Execute the command (a string) in a subshell."""
     args = _cmdline2listimpl(command)
     args = _escape_args(args)
+    if _shell_command is None:
+        raise OSError("Unable to locate or access a valid system shell to execute system() command.")
     args = _shell_command + args
     cwd = os.getcwdu()
 


Lib/test/test_subprocess.py
```diff
@@ -724,10 +724,46 @@ def test_invalid_args(self):
                                "-c", 1])
 
 
+class SecurityManagerTestCase(unittest.TestCase):
+    def test_subprocess_shell_with_security_manager(self):
+        if not jython:
+            return
+
+        import java.lang.SecurityManager as SecurityManager
```
Stewori, line 732: Just a note: [SecurityManager is deprecated for removal](https://apidia.net/java/OpenJDK/26/?pck=java.lang&cls=.SecurityManager). Perhaps, it's premature to address this here, before it's actually removed. Once it's removed and the definitive JDK version of the removal is known, we can add a conditional skip.

Lib/test/test_subprocess.py
```diff
@@ -724,10 +724,46 @@ def test_invalid_args(self):
                                "-c", 1])
 
 
+class SecurityManagerTestCase(unittest.TestCase):
+    def test_subprocess_shell_with_security_manager(self):
+        if not jython:
```
jeff5, line 729: A nicer way to do this is the decorator `@unittest.skipIf(not jython, ...`. There's a `@unittest.skipIf(jython` somewhere in the file. It could be applied to the `class`.

Really appreciate you adding a test.

Lib/test/test_subprocess.py
```diff
@@ -724,10 +724,46 @@ def test_invalid_args(self):
...
+            return
+
+        import java.lang.SecurityManager as SecurityManager
+        import java.lang.System as System
+        import java.security.AccessControlException as AccessControlException
+
+        class MySecurityManager(SecurityManager):
+            def checkRead(self, file, context=None):
+                if file == "/bin/sh" or file == "cmd.exe" or file.endswith("sh"):
```
jeff5, line 738: This needs a little more thought. Ok, it's only for the test and we're not proposing a real security manager. But it fails on my Windows machine because `file` contains `"C:\Windows\System32\cmd.exe"`. And what is the intent of `endswith("sh")` anyway?  So, probably `endswith("cmd.exe")`.

It passes on the Windows CI because we do not enable the `subprocess` resource. (Seems like an omission right now.)

Fixed like that, and with `checkRead` printing what `file` it has be asked about, I find that `subprocess._setup_platform()` involves looking for an awful lot of files.

Lib/test/test_subprocess.py
```diff
@@ -724,10 +724,46 @@ def test_invalid_args(self):
                                "-c", 1])
 
 
+class SecurityManagerTestCase(unittest.TestCase):
```
jeff5, line 727: Or on second thoughts, is this maybe happier in `test_subprocess_jy.py`, where everything is Jython-specific?

jeff5 (changes requested):
This seems like a valid bug and a sensible solution, thanks. I've only found points to criticise in the test. But thanks for adding one.

PR #409, opened 2025-12-20
Upgrade jffi to version 1.3.14 with JDK23+ support

wfouche:
Resolves the  following warning with Jython is used with JDK23 or later.

```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::putLong has been called by com.kenai.jffi.UnsafeMemoryIO$UnsafeMemoryIO64 (file:/home/wfouche/.m2/repository/com/github/jnr/jffi/1.3.13/jffi-1.3.13.jar)
WARNING: Please consider reporting this to the maintainers of class com.kenai.jffi.UnsafeMemoryIO$UnsafeMemoryIO64
WARNING: sun.misc.Unsafe::putLong will be removed in a future release
```

To be considered for inclusion in Jython 2.7.5

PR #424, opened 2026-05-09
Support Java 25 runtime compatibility

### LINKED ISSUE #413: Jython 2.7.5 release with Java 25 support
The following three PRs ensure that Jython works with Java 25:

- https://github.com/jython/jython/pull/410
- https://github.com/jython/jython/pull/409
- https://github.com/jython/jython/pull/407

robertpatrick:
Fixes #413 

## Summary

This PR updates Jython for runtime compatibility with JDK 25 while keeping the release built with the lowest supported JDK target. The release artifacts were built with JDK 8 and then tested on JDK 8, 11, 17, 21, and 25 from the same build output. No multi-release JAR was needed.

This builds on the already-merged jffi update from #409 and incorporates/adapts fixes from:

- #407 for `stdout.encoding` / `sun.stdout.encoding` handling on newer JDKs
- #410 for JDK 25 native-access and module-open handling

## Changes

- Add JDK 25 runtime launcher support:
  - `--enable-native-access=ALL-UNNAMED`
  - `--sun-misc-unsafe-memory-access=allow`
- Add JDK 9+ launcher/module opens needed by current runtime dependencies:
  - `java.base/java.io`
  - `java.base/sun.nio.ch`
- Add matching manifest attributes to runtime/test jars:
  - `Enable-Native-Access: ALL-UNNAMED`
  - `Add-Opens: java.base/java.io java.base/sun.nio.ch`
- Update ASM from 9.7 to 9.9.1.
- Add `jython.test.java.opts` so CI/test runs can pass Java proxy/system properties without editing build files.
-  Added a `jython.test.socket.unreachableHost` system property to override the test_socket unreachable-host timeout target while preserving the existing `192.0.2.42` default.
- Added new `test-built` Ant target that runs the tests without calling developer-build
- Adjust version-sensitive tests for behavior changes in newer JDKs.
- Skip SecurityManager and class-unloading leak tests where they are no longer valid on JDK 25.
- Improve proxy-aware network tests while preserving direct localhost behavior.
- Make `modjy` test failures propagate as process failures.
- Suppress test-only noisy Netty logging from deliberate httplib socket error paths.
- Release any partially consumed Netty socket buffer held in incoming_head during socket close.
- Make `test_mailbox.test_clean` robust against stale deterministic tmp/foo and tmp/bar files.
- Copy `profile.properties` into `dist/javalib` and adjust the bundled JIP profiler properties so the Java 8/11 `--profile` launcher tests produce output without noisy agent stack traces.
- On Java 13+, report --profile as unsupported because the bundled Java Interactive Profiler agent is not compatible with modern class verification.

## Proxy-sensitive network tests

When an external HTTP proxy is configured, this PR skips only the network cases that cannot work through an HTTP proxy:

- `test_urllib2net` external FTP test is skipped when an external HTTP proxy is configured because it requires direct FTP access. Neither FTP-over-HTTP-proxy support nor Java FTP proxy property support is implemented in this PR.
- `test_ssl_jy` raw external SSL socket tests are skipped because they open raw sockets and cannot use HTTP CONNECT.
- `test_smtpnet` raw external SMTP SSL tests are skipped because they require direct SMTP-over-SSL socket access.

Higher-level external HTTPS tests were adjusted to use CONNECT where appropriate. Localhost/direct tests continue to run directly.

The relevant generic configuration surfaces for proxied test environments are:

- Environment variables: http_proxy, https_proxy, no_proxy.
- Java system properties passed through `jython.test.java.opts` system property: 
  - `http.proxyHost`
  - `http.proxyPort`
  - `http.nonProxyHosts`
  - `https.proxyHost`
  - `https.proxyPort`
  - `https.nonProxyHosts`.

## Validation

Release artifacts were built with JDK 8 using `ant clean full-build`:

- `dist/jython.jar`
- `dist/jython-standalone.jar`
- `dist/jython-installer.jar`

The resulting artifacts still target JDK 8 bytecode/release metadata.

The regression matrix was run with:

- JDK 8: `ant test`
- JDK 11: `ant test-built`
- JDK 17: `ant test-built`
- JDK 21: `ant test-built`
- JDK 25: `ant test-built`

Results:

- JDK 8: 388 tests OK, 2 tests skipped; modjy OK (104 tests); BUILD SUCCESSFUL.
- JDK 11: 388 tests OK, 2 tests skipped; modjy OK (104 tests); BUILD SUCCESSFUL.
- JDK 17: 388 tests OK, 2 tests skipped; modjy OK (104 tests); BUILD SUCCESSFUL.
- JDK 21: 388 tests OK, 2 tests skipped; modjy OK (104 tests); BUILD SUCCESSFUL.
- JDK 25: 388 tests OK, 2 tests skipped; modjy OK (104 tests); BUILD SUCCESSFUL.

The two top-level skipped tests are existing resource-gated tests:

- `test_codecmaps_hk`, requires `urlfetch`
- `test_curses`, requires `curses`

They are not JDK 25 failures and are not proxy-related.

Additional focused checks used during development included JDK 25 command-line behavior, float behavior, imports, Java integration, subprocess/threading behavior, HTTP/URL handling, warning suppression, `test_mailbox` cleanup, and JDK 21/JDK 25 socket/SSL behavior.


ohumbel:
This sounds promising!

On which platform, and with which exact command did you run the regression tests?
I'd like to reproduce them.

robertpatrick:
I ran this on my Mac laptop.  Are there additional tests other than the CI tests Codex ran?  Happy to run those if you point me in the right direction...

`ant init regrtest-unix-ci`

robertpatrick:
@ohumbel I expanded the scope of the PR.  I ran the following commands:

- JDK8: `ant clean full-build` and `ant test`
- JDK11: `ant test-built`
- JDK17: `ant test-built`
- JDK21: `ant test-built`
- JDK25: `ant test-built`

Let me know if there is anything else I can do to help.

ohumbel:
> @ohumbel I expanded the scope of the PR. I ran the following commands:
> 
>     * JDK8: `ant clean full-build` and `ant test`

@robertpatrick When running `ant test`, I get:
```
launchertest-built:
     [exec] + '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.3PA6YJJSWP/directory with spaces/bin/jython' -J-version
     [exec] ++ egrep -c '^java version '
     [exec] + '[' 0 == 1 ']'

BUILD FAILED
/Users/oti/stuff/gitrepo/robertpatrick/jython/build.xml:1493: exec returned: 1

Total time: 1 minute 15 seconds
```
The same error both on Mac and Ubuntu, using Java 8: 
```
java -version 
openjdk version "1.8.0_492"
OpenJDK Runtime Environment (Temurin)(build 1.8.0_492-b09)
OpenJDK 64-Bit Server VM (Temurin)(build 25.492-b09, mixed mode)
```

robertpatrick:
@ohumbel Sorry, I don't use Open JDK.  PR updated, please try again.

jeff5:
@robertpatrick This all looks relevant and helpful. Thanks for tracking down so many apparent version-dependent changes. The additions to `build.xml` so that we can run a built copy are helpful too.

I know you've contributed before, but I can't trace a record of signing the [CLA](https://github.com/jython/jython/blob/master/ProcessDocs/contributor_agreement.md). It may be just that when you did, it wasn't correlated with a GitHub id. Or it may be we weren't all that strict.

This is really too many things in one PR. To help us know which changes are related, as reviewers or when someone does the archaeology later to understand a change they find problematic, could we divide this up please? Two of them @wfouche has cooking as PRs for anyway, as you identify, so if we can bank those we need not include them here.

ohumbel:
> @ohumbel Sorry, I don't use Open JDK. PR updated, please try again.

@robertpatrick If you have a look at [ant-regrtest.yml](https://github.com/jython/jython/blob/master/.github/workflows/ant-regrtest.yml), you can see that the pipelines also use the Eclipse Temurin distributions. That's the reason I test with those.

I have a positive and negative feedback for you.
On the positive side, I can reproduce your test results *if I comment out `launchertest`*.
On macOS Tahoe 26.4.1 (Intel), with the following java versions:
```
sdk use java 8.0.492-tem
sdk use java 11.0.31-tem
sdk use java 17.0.19-tem
sdk use java 21.0.11-tem
sdk use java 25.0.3-tem
```
I consistently get
```
     [exec] 388 tests OK.
     [exec] 2 tests skipped:
```
Thats a good sign for me.
I need some more time to do the same runs on Ubuntu, and possibly on Windows.

`launchertest` gets a bit further (`sdk use java 8.0.492-tem`), but fails as follows:
```
launchertest-built:
     [exec] + '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/directory with spaces/bin/jython' -J-version
     [exec] ++ egrep -c '^(java|openjdk) version '
     [exec] + '[' 1 == 1 ']'
     [exec] + '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/directory with spaces/bin/jython' --version
     [exec] ++ egrep -c '^Jython '
     [exec] + '[' 1 == 1 ']'
     [exec] + JYTHON_OPTS=--version
     [exec] + '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/directory with spaces/bin/jython'
     [exec] ++ egrep -c '^Jython '
     [exec] + '[' 1 == 1 ']'
     [exec] + '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/directory with spaces/bin/jython' --help
     [exec] ++ egrep -c '^usage: '
     [exec] + '[' 1 == 1 ']'
     [exec] + '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/directory with spaces/bin/jython' --help
     [exec] ++ egrep -c '^Jython launcher options:'
     [exec] + '[' 1 == 1 ']'
     [exec] + set +e
     [exec] + '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/directory with spaces/bin/jython' -c 'import sys; sys.exit(5)'
     [exec] + '[' 5 == 5 ']'
     [exec] + set -e
     [exec] ++ '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/directory with spaces/bin/jython' -c 'import sys; print sys.executable is not None,'
     [exec] + '[' True == True ']'
     [exec] ++ java_major_version '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/directory with spaces/bin/jython'
     [exec] ++ local output version major
     [exec] +++ '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/directory with spaces/bin/jython' -J-version
     [exec] ++ output='openjdk version "1.8.0_492"
     [exec] OpenJDK Runtime Environment (Temurin)(build 1.8.0_492-b09)
     [exec] OpenJDK 64-Bit Server VM (Temurin)(build 25.492-b09, mixed mode)'
     [exec] +++ expr 'openjdk version "1.8.0_492"
     [exec] OpenJDK Runtime Environment (Temurin)(build 1.8.0_492-b09)
     [exec] OpenJDK 64-Bit Server VM (Temurin)(build 25.492-b09, mixed mode)' : '.*version "\([^"]*\)"'
     [exec] ++ version=1.8.0_492
     [exec] ++ case "$version" in
     [exec] ++ major=8.0_492
     [exec] ++ major=8
     [exec] ++ echo 8
     [exec] + JAVA_MAJOR=8
     [exec] + JDB_OUTPUT=/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/jdb.out
     [exec] + set +e
     [exec] + echo run
     [exec] + sleep 3
     [exec] + '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/directory with spaces/bin/jython' --jdb -c 'print '\''\ntest'\'''
     [exec] + JDB_STATUS=0
     [exec] + set -e
     [exec] ++ egrep -c '^test' /var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/jdb.out
     [exec] + '[' 0 == 1 ']'
     [exec] + cat /var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/jdb.out
     [exec] Initializing jdb ...
     [exec] > run org.python.util.jython -c "print '\ntest'"
     [exec] VM start exception: VM initialization failed for: /Users/oti/.sdkman/candidates/java/8.0.492-tem/jre/bin/java -Xmx512m -Xss2560k -Dfile.encoding=UTF-8 -classpath /var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/directory with spaces/jython-dev.jar:/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/directory with spaces/jython-test.jar:/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/directory with spaces/javalib/*: -Dpython.home=/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/directory with spaces -Dpython.executable=/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.zuy6atO8ey/directory with spaces/bin/jython -Xdebug -Xrunjdwp:transport=dt_socket,address=rope.local:57105,suspend=y org.python.util.jython -c print '\ntest'
     [exec] 
     [exec] Error: Could not find or load main class with
     [exec] 
     [exec] Fatal error:
     [exec] Target VM failed to initialize.
     [exec] + '[' 0 -ne 0 ']'
     [exec] + exit 1

BUILD FAILED
/Users/oti/stuff/gitrepo/robertpatrick/jython/build.xml:1493: exec returned: 1

Total time: 1 minute 23 seconds
```

robertpatrick:
@jeff5 

>I know you've contributed before, but I can't trace a record of signing the [CLA](https://github.com/jython/jython/blob/master/ProcessDocs/contributor_agreement.md). It may be just that when you did, it wasn't correlated with a GitHub id. Or it may be we weren't all that strict.

Done

> This is really too many things in one PR. 

Well...  It is hard to know if a PR is useful if you cannot actually build and test it with the stated goal.  

> To help us know which changes are related, as reviewers or when someone does the archaeology later to understand a change they find problematic, could we divide this up please?

And how exactly would you suggest we do that?  Send a bunch of related PRs, none of which actually achieve the stated goal?  How exactly do you vet those PRs individually?

> Two of them @wfouche has cooking as PRs for anyway, as you identify, so if we can bank those we need not include them here.

Again, I need Jython support for JDK 25 and so I achieved this and met all of the goals that have been stated.  I did give credit to those two PRs but if they were ready, why haven't you merged them already?  I seem to recall that you had some feedback on one of them that they did not achieve the objective the way that you wanted.

robertpatrick:
@ohumbel Updated the PR to address the issue with the launchertest-built target.

> I consistently get

     [exec] 388 tests OK.
     [exec] 2 tests skipped:

Those two skips are from the existing Ant/regrtest resource configuration, not from this PR.

Both `ant test` and the upstream `regrtest-ci` target run regrtest with:

```
    --use network,subprocess
```

They do not enable `urlfetch` or `curses`.

- `test_codecmaps_hk` requires `urlfetch` because it downloads an external codec mapping file. We can run that as an additional optional validation where outbound HTTP is available, but I would not make the main JDK compatibility matrix depend on a third-party URL.

- `test_curses` requires the `curses` resource and then imports the native `_curses` module. Jython does not ship a `_curses` implementation, and installing system ncurses/libcurses does not make CPython’s native extension available to Jython. Making that test run would be a separate Jython curses-support feature, not a JDK 25 compatibility issue.

I am happy to go do more work to make sure that all of your requirements for OS and JDK versions are met.  However, I think you and @jeff5 need to decide what you want to do.  I do not want to waste time.

We need a working version of Jython that runs properly on JDK25, and we need it soon.  I tried to help the process along with this PR but apparently, didn't do it the way @jeff5 wanted.  So, what is the plan for getting JDK 25 support added in the near future if it is not this PR?

Note that I can continue my work on this PR to verify OS/JDK support.  However, if you two decide not to merge it after I satisfy the OS/JDK requirements, we will be forced to fork Jython, apply the PR, and move forward.  As one of the largest consumers of Jython, that would be bad for Jython as a whole, and create extra work for us but we have dates to meet.

ohumbel:
@robertpatrick 
The 2 skips are expected and completely ok, I think.

Being "only" a contributor like you, I just wanted to help verify the PR locally - time is always short on all sides.
My plans are to help moving Jython forward, towards modern Java and up to date external libraries.

ohumbel:
@robertpatrick 
`launchertest` seems to be a hard nut to crack. I ran this target only: `ant launchertest`:
```
launchertest-built:
     [exec] + '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.iA57H96DUy/directory with spaces/bin/jython' -J-version
     [exec] ++ egrep -c '^(java|openjdk) version '
     [exec] + '[' 1 == 1 ']'
     [exec] + '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.iA57H96DUy/directory with spaces/bin/jython' --version
     [exec] ++ egrep -c '^Jython '
     [exec] + '[' 1 == 1 ']'
     [exec] + JYTHON_OPTS=--version
     [exec] + '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.iA57H96DUy/directory with spaces/bin/jython'
     [exec] ++ egrep -c '^Jython '
     [exec] + '[' 1 == 1 ']'
     [exec] + '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.iA57H96DUy/directory with spaces/bin/jython' --help
     [exec] ++ egrep -c '^usage: '
     [exec] + '[' 1 == 1 ']'
     [exec] + '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.iA57H96DUy/directory with spaces/bin/jython' --help
     [exec] ++ egrep -c '^Jython launcher options:'
     [exec] + '[' 1 == 1 ']'
     [exec] + set +e
     [exec] + '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.iA57H96DUy/directory with spaces/bin/jython' -c 'import sys; sys.exit(5)'
     [exec] + '[' 5 == 5 ']'
     [exec] + set -e
     [exec] ++ '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.iA57H96DUy/directory with spaces/bin/jython' -c 'import sys; print sys.executable is not None,'
     [exec] + '[' True == True ']'
     [exec] ++ java_major_version '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.iA57H96DUy/directory with spaces/bin/jython'
     [exec] ++ local output version major
     [exec] +++ '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.iA57H96DUy/directory with spaces/bin/jython' -J-version
     [exec] ++ output='openjdk version "1.8.0_492"
     [exec] OpenJDK Runtime Environment (Temurin)(build 1.8.0_492-b09)
     [exec] OpenJDK 64-Bit Server VM (Temurin)(build 25.492-b09, mixed mode)'
     [exec] +++ expr 'openjdk version "1.8.0_492"
     [exec] OpenJDK Runtime Environment (Temurin)(build 1.8.0_492-b09)
     [exec] OpenJDK 64-Bit Server VM (Temurin)(build 25.492-b09, mixed mode)' : '.*version "\([^"]*\)"'
     [exec] ++ version=1.8.0_492
     [exec] ++ case "$version" in
     [exec] ++ major=8.0_492
     [exec] ++ major=8
     [exec] ++ echo 8
     [exec] + JAVA_MAJOR=8
     [exec] + JDB_OUTPUT=/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.iA57H96DUy/jdb.out
     [exec] + set +e
     [exec] + echo run
     [exec] + sleep 3
     [exec] + '/var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.iA57H96DUy/directory with spaces/bin/jython' --jdb -c 'print '\''\ntest'\'''
     [exec] + JDB_STATUS=0
     [exec] + set -e
     [exec] ++ egrep -c '^test' /var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.iA57H96DUy/jdb.out
     [exec] + '[' 0 == 1 ']'
     [exec] + cat /var/folders/g9/t4tbc7x57jn74ch4g1978c3w0000gn/T/test-jython.XXXXXX.iA57H96DUy/jdb.out
     [exec] Initializing jdb ...
     [exec] > run org.python.util.jython -c "print '\ntest'"
     [exec] Set uncaught java.lang.Throwable
     [exec] Set deferred uncaught java.lang.Throwable
     [exec] > 
     [exec] VM Started: Input stream closed.
     [exec] + '[' 0 -ne 0 ']'
     [exec] + exit 1

BUILD FAILED
/Users/oti/stuff/gitrepo/robertpatrick/jython/build.xml:1493: exec returned: 1

Total time: 17 seconds
```

Stewori:
> However, if you two decide not to merge it after I satisfy the OS/JDK requirements, we will be forced to fork Jython, apply the PR, and move forward. As one of the largest consumers of Jython, that would be bad for Jython as a whole, and create extra work for us but we have dates to meet.

Please, let's remind ourselves that everyone is working on this voluntarily. No need for "threats" like this - I really see no way how Jython 2.7.5 could be released *without* including this PR the one or other way. So, no worries - it will surely be accepted once remaining issues are solved.

That said, if we merge @wfouche 's PRs first, the diff should just adopt and we can merge this PR on top, I suppose. (Perhaps it will require resolving a mild merge conflict if at all.)

robertpatrick:
@ohumbel I will go download and install these non-standard JDK distributions and try to resolve the issue.

jeff5:
CLA: thank-you.

> > ... could we divide this up please?
> 
> And how exactly would you suggest we do that? Send a bunch of related PRs, none of which actually achieve the stated goal? How exactly do you vet those PRs individually?

I think the obvious approach is to address test failures (in `regrtest` with various `-u` settings), in groups that have a common cause. #407 and #410 achieve isolated advances that can be reviewed, and tweaks proposed, although #413 overlooks what may be left to do.

robertpatrick:
@jeff5 I have no time or interest in dividing this up.  

Just so we are clear, I am doing this work because my employer needs these changes and we want to be good open-source citizens.  I was not trying to make threats earlier, I was stating the facts that we have to ship this and have dates that we need to meet.

I will continue to focus on testing and resolving issues in their non-standard JDKs and Linux/Windows.  If you want to break it up, into separate PRs, be my guest.  For me, this is not about credit, my ego, or any such thing.  It is about getting the changes made as soon as possible to meet the dates my employer has.

robertpatrick:
@ohumbel I cannot reproduce your issue using the same Eclipse Temurin JDK 1.8.0_492 on macOS Tahoe 26.4.1.  Same result without the -D overrides for the JDBC jars.

```
rpatrick@rpatrick-mac1 jython % java -version
openjdk version "1.8.0_492"
OpenJDK Runtime Environment (Temurin)(build 1.8.0_492-b09)
OpenJDK 64-Bit Server VM (Temurin)(build 25.492-b09, mixed mode)
rpatrick@rpatrick-mac1 jython %  ant launchertest \
    -Dinformix.jar=/tmp/jdbc-4.10.12.jar \
    -Doracle.jar=/opt/weblogic/soa12214/oracle_common/modules/oracle.jdbc/ojdbc8.jar
Buildfile: /Users/rpatrick/Projects/jython/build.xml

developer-preinit:

common-dirs:

common-jars:

common-paths:

common-constants:

common-config:

common-version-strings:

init:

antlr-up-to-date:

antlr-gen:

git-present:

git-is-tagged:

git-is-unmodified:

brand-up-to-date:

brand-version-metadata:

brand-version-nogit:

brand-version:
     [echo] Writing git and build metadata to version.properties.
     [echo] jython.version        = 2.7.5a1-DEV
     [echo] build.git.branch      = remotes/origin/gh-413-jdk25-compat
     [echo] build.git.version     = 43bc8e8bb
     [echo] build.git.tag         = heads/gh-413-jdk25-compat

compile:

expose:

jar:
      [jar] Updating jar: /Users/rpatrick/Projects/jython/dist/jython-dev.jar

compile-test:

jar-test:
      [jar] Updating jar: /Users/rpatrick/Projects/jython/dist/jython-test.jar

copy-javalib:

copy-bin:

copy-cpythonlib:

copy-lib:

copy-dev:

pycompile:

developer-build:

launchertest-built:
     [exec] + '/var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/directory with spaces/bin/jython' -J-version
     [exec] ++ egrep -c '^(java|openjdk) version '
     [exec] + '[' 1 == 1 ']'
     [exec] + '/var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/directory with spaces/bin/jython' --version
     [exec] ++ egrep -c '^Jython '
     [exec] + '[' 1 == 1 ']'
     [exec] + JYTHON_OPTS=--version
     [exec] + '/var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/directory with spaces/bin/jython'
     [exec] ++ egrep -c '^Jython '
     [exec] + '[' 1 == 1 ']'
     [exec] + '/var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/directory with spaces/bin/jython' --help
     [exec] ++ egrep -c '^usage: '
     [exec] + '[' 1 == 1 ']'
     [exec] + '/var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/directory with spaces/bin/jython' --help
     [exec] ++ egrep -c '^Jython launcher options:'
     [exec] + '[' 1 == 1 ']'
     [exec] + set +e
     [exec] + '/var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/directory with spaces/bin/jython' -c 'import sys; sys.exit(5)'
     [exec] + '[' 5 == 5 ']'
     [exec] + set -e
     [exec] ++ '/var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/directory with spaces/bin/jython' -c 'import sys; print sys.executable is not None,'
     [exec] + '[' True == True ']'
     [exec] ++ java_major_version '/var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/directory with spaces/bin/jython'
     [exec] ++ local output version major
     [exec] +++ '/var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/directory with spaces/bin/jython' -J-version
     [exec] ++ output='openjdk version "1.8.0_492"
     [exec] OpenJDK Runtime Environment (Temurin)(build 1.8.0_492-b09)
     [exec] OpenJDK 64-Bit Server VM (Temurin)(build 25.492-b09, mixed mode)'
     [exec] +++ expr 'openjdk version "1.8.0_492"
     [exec] OpenJDK Runtime Environment (Temurin)(build 1.8.0_492-b09)
     [exec] OpenJDK 64-Bit Server VM (Temurin)(build 25.492-b09, mixed mode)' : '.*version "\([^"]*\)"'
     [exec] ++ version=1.8.0_492
     [exec] ++ case "$version" in
     [exec] ++ major=8.0_492
     [exec] ++ major=8
     [exec] ++ echo 8
     [exec] + JAVA_MAJOR=8
     [exec] + JDB_OUTPUT=/var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/jdb.out
     [exec] + set +e
     [exec] + echo run
     [exec] + sleep 3
     [exec] + '/var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/directory with spaces/bin/jython' --jdb -c 'print '\''\ntest'\'''
     [exec] + JDB_STATUS=0
     [exec] + set -e
     [exec] ++ egrep -c '^test' /var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/jdb.out
     [exec] + '[' 1 == 1 ']'
     [exec] + :
     [exec] + '[' 8 -lt 13 ']'
     [exec] + '/var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/directory with spaces/bin/jython' --profile -c pass
     [exec] ++ egrep -c '^\| Most expensive methods'
     [exec] + '[' 2 -ge 1 ']'
     [exec] + '[' -f profile.txt ']'
     [exec] + rm profile.txt
     [exec] + CLASSPATH='/var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/directory with spaces/Lib/test/blob.jar'
     [exec] + '/var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/directory with spaces/bin/jython' -c 'print __import__('\''Blob'\'')'
     [exec] ++ egrep -c Blob
     [exec] + '[' 1 == 1 ']'
     [exec] + '[' 8 -lt 13 ']'
     [exec] + CLASSPATH='/var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/directory with spaces/Lib/test/blob.jar'
     [exec] + '/var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/directory with spaces/bin/jython' --profile -c 'print __import__('\''Blob'\'')'
     [exec] ++ egrep -c Blob
     [exec] Using the generic class loader filter.
     [exec] Java Interactive Profiler: starting
     [exec] Controller -- shuttingdown
     [exec] + '[' 1 == 1 ']'
     [exec] + set +ex
     [exec] + /Users/rpatrick/Projects/jython/dist/bin/jython -J-version
     [exec] ++ egrep -c '^(java|openjdk) version '
     [exec] + '[' 1 == 1 ']'
     [exec] + /Users/rpatrick/Projects/jython/dist/bin/jython --version
     [exec] ++ egrep -c '^Jython '
     [exec] + '[' 1 == 1 ']'
     [exec] + JYTHON_OPTS=--version
     [exec] + /Users/rpatrick/Projects/jython/dist/bin/jython
     [exec] ++ egrep -c '^Jython '
     [exec] + '[' 1 == 1 ']'
     [exec] + /Users/rpatrick/Projects/jython/dist/bin/jython --help
     [exec] ++ egrep -c '^usage: '
     [exec] + '[' 1 == 1 ']'
     [exec] + /Users/rpatrick/Projects/jython/dist/bin/jython --help
     [exec] ++ egrep -c '^Jython launcher options:'
     [exec] + '[' 1 == 1 ']'
     [exec] + set +e
     [exec] + /Users/rpatrick/Projects/jython/dist/bin/jython -c 'import sys; sys.exit(5)'
     [exec] + '[' 5 == 5 ']'
     [exec] + set -e
     [exec] ++ /Users/rpatrick/Projects/jython/dist/bin/jython -c 'import sys; print sys.executable is not None,'
     [exec] + '[' True == True ']'
     [exec] ++ java_major_version /Users/rpatrick/Projects/jython/dist/bin/jython
     [exec] ++ local output version major
     [exec] +++ /Users/rpatrick/Projects/jython/dist/bin/jython -J-version
     [exec] ++ output='openjdk version "1.8.0_492"
     [exec] OpenJDK Runtime Environment (Temurin)(build 1.8.0_492-b09)
     [exec] OpenJDK 64-Bit Server VM (Temurin)(build 25.492-b09, mixed mode)'
     [exec] +++ expr 'openjdk version "1.8.0_492"
     [exec] OpenJDK Runtime Environment (Temurin)(build 1.8.0_492-b09)
     [exec] OpenJDK 64-Bit Server VM (Temurin)(build 25.492-b09, mixed mode)' : '.*version "\([^"]*\)"'
     [exec] ++ version=1.8.0_492
     [exec] ++ case "$version" in
     [exec] ++ major=8.0_492
     [exec] ++ major=8
     [exec] ++ echo 8
     [exec] + JAVA_MAJOR=8
     [exec] + JDB_OUTPUT=/var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/jdb.out
     [exec] + set +e
     [exec] + echo run
     [exec] + sleep 3
     [exec] + /Users/rpatrick/Projects/jython/dist/bin/jython --jdb -c 'print '\''\ntest'\'''
     [exec] + JDB_STATUS=0
     [exec] + set -e
     [exec] ++ egrep -c '^test' /var/folders/lq/3_tv3hpd3yv2f7r2hf4yzbfc0000gp/T/test-jython.XXXXXX.MdK0j8mmXi/jdb.out
     [exec] + '[' 1 == 1 ']'
     [exec] + :
     [exec] + '[' 8 -lt 13 ']'
     [exec] + /Users/rpatrick/Projects/jython/dist/bin/jython --profile -c pass
     [exec] ++ egrep -c '^\| Most expensive methods'
     [exec] + '[' 2 -ge 1 ']'
     [exec] + '[' -f profile.txt ']'
     [exec] + rm profile.txt
     [exec] + CLASSPATH=/Users/rpatrick/Projects/jython/dist/Lib/test/blob.jar
     [exec] + /Users/rpatrick/Projects/jython/dist/bin/jython -c 'print __import__('\''Blob'\'')'
     [exec] ++ egrep -c Blob
     [exec] + '[' 1 == 1 ']'
     [exec] + '[' 8 -lt 13 ']'
     [exec] + CLASSPATH=/Users/rpatrick/Projects/jython/dist/Lib/test/blob.jar
     [exec] + /Users/rpatrick/Projects/jython/dist/bin/jython --profile -c 'print __import__('\''Blob'\'')'
     [exec] ++ egrep -c Blob
     [exec] Using the generic class loader filter.
     [exec] Java Interactive Profiler: starting
     [exec] Controller -- shuttingdown
     [exec] + '[' 1 == 1 ']'
     [exec] + set +ex
     [exec] Test successful.

launchertest:

BUILD SUCCESSFUL
Total time: 43 seconds
rpatrick@rpatrick-mac1 jython % 
```

robertpatrick:
Tested on Oracle Linux 9.6.  The only update that was required was adding a system property that allowed me to override the test_socket.py file's  default, hard-coded unreachable host `192.0.2.42`--which was reachable on our network.

The test matrix run was:

- Eclipse Temurin JDK 8u492: PASS, 388 tests OK, 2 skipped, modjy OK (104 tests)
- Eclipse Temurin JDK 11.0.31: PASS, 388 tests OK, 2 skipped, modjy OK (104 tests)
- Eclipse Temurin JDK 17.0.19: PASS, 388 tests OK, 2 skipped, modjy OK (104 tests)
- Eclipse Temurin JDK 21.0.11: PASS, 388 tests OK, 2 skipped, modjy OK (104 tests)
- Eclipse Temurin JDK 25.0.3: PASS, 388 tests OK, 2 skipped, modjy OK (104 tests)

I will go test on Windows next.

Lib/test/test_httplib.py
```diff
@@ -851,7 +872,7 @@ def test_networked_trusted_by_default_cert(self):
         # Default settings: requires a valid cert from a trusted CA
         test_support.requires('network')
         with test_support.transient_internet('www.python.org'):
-            h = httplib.HTTPSConnection('www.python.org', 443)
+            h = _external_https_connection('www.python.org', 443)
```
Stewori, line 876: Later in testPythonOrg, you replace www.python.org with www.pythontest.net. Can this be done here, too? I mean, using www.pythontest.net instead of www.python.org would then be consistent among tests.

Lib/test/test_socket.py
```diff
@@ -19,6 +19,8 @@
 
 PORT = 50100
 HOST = 'localhost'
+UNREACHABLE_HOST = java.lang.System.getProperty(
+    'jython.test.socket.unreachableHost', '192.0.2.42')
```
Stewori, line 23: How does this variable instead of hard-coded '192.0.2.42' solve any Java compatibility issue? Is there a reason to expect the system property "jython.test.socket.unreachableHost" to be ever populated? It looks just like refactoring to me. Are the edits to test_socket.py actually necessary?

Lib/test/test_urllib2.py
```diff
@@ -1490,14 +1490,22 @@ def test_HTTPError_interface_call(self):
...
     from test import test_urllib2
-    test_support.run_doctest(test_urllib2, verbose)
-    test_support.run_doctest(urllib2, verbose)
-    tests = (TrivialTests,
-             OpenerDirectorTests,
-             HandlerTests,
-             MiscTests,
-             RequestTests)
-    test_support.run_unittest(*tests)
+    old_opener = urllib2._opener
```
Stewori: Can you explain why _opener has to be backupped like this? How does it get possibly overwritten so it would need restore? Why was this not needed before? I don't see how the other change below (unsetting some environment variables) would impact _opener.

Lib/test/test_urllibnet.py
```diff
@@ -11,6 +11,8 @@
 
 mimetools = test_support.import_module("mimetools", deprecated=True)
 
+URL = test_support.TEST_HTTP_URL
```
Stewori, line 15: Just out of curiosity, what url is this? Maybe it would be better to have consistently one test url in all tests. We already have www.python.org and www.pythontest.net in other tests. Why not always the same, perhaps always this one?

Stewori: Ah, okay, is it because a http url is needed rather than https? And python.org is https-only nowadays. Perhaps clarify this in a short comment.

Lib/test/test_urllibnet.py
```diff
@@ -95,11 +103,6 @@ def test_info(self):
     def test_geturl(self):
         # Make sure same URL as opened is returned by geturl.
         #
-        # This test has been changed from what's currently in our
-        # lib-python/2.7 for Jython due to recent updates at the
-        # python.org to use https; other tests can take advantate of
-        # URL redirection
-        URL = "https://www.python.org/"
         open_url = self.urlopen(URL)
```
Stewori, line 108: Okay, for this test the change is real. Previously it *was* tested with the https url (https://www.python.org), now URL would point to the http (not https any more) url from test_support. Is it irrelevant for this urlopen call whether it's a http or https url?

Lib/test/test_urllibnet.py
```diff
@@ -189,14 +193,13 @@ def test_specified_path(self):
...
-        file_location, header = self.urlretrieve("http://www.python.org/")
+        file_location, header = self.urlretrieve(URL)
         os.unlink(file_location)
         self.assertIsInstance(header, mimetools.Message,
                               "header is not an instance of mimetools.Message")
 
     def test_data_header(self):
-        logo = "http://www.python.org/community/logos/python-logo-master-v3-TM.png"
-        file_location, fileheaders = self.urlretrieve(logo)
+        file_location, fileheaders = self.urlretrieve(URL)
```
Stewori, line 206: This is still the test http url from test_support, right? How can we expect it to point to an actual file? Can you add a comment which file it points to, actually?

src/org/python/core/PySystemState.java
```diff
@@ -1020,16 +1020,29 @@ private static void initRegistry(Properties preProperties, Properties postProper
      * @param props in which to look for clues (normally the Jython registry)
      * @return the console encoding (and never {@code null})
      */
-    private static String getConsoleEncoding(Properties props) {
+    static String getConsoleEncoding(Properties props) {
```
Stewori, line 1023: Okay, I suppose the changes to this file are those from Werner's PR. This file diff will update once we merge that one. We'll see how this turns out.

src/org/python/util/jython.java
```diff
@@ -235,7 +235,7 @@ public static void loggingToConsole() {
         }
 
         if (exception != null) {
-            logger.log(Level.WARNING, "Unable to format console messages: {0}",
+            logger.log(Level.FINE, "Unable to format console messages: {0}",
```
Stewori, line 238: Is this relevant for a potential doctest failure? Or is it just silencing some noise?

src/shell/jython
```diff
@@ -244,27 +343,36 @@ fi
...
-"${JAVA_CMD[@]}" $JAVA_OPTS "${java_args[@]}" -Dpython.home="$JYTHON_HOME" \
-  -Dpython.executable="$PRG" org.python.util.jython $JYTHON_OPTS "$@"
+"${JAVA_CMD[@]}" $JAVA_OPTS "${java_args[@]}" -Dpython.home="$JYTHON_HOME_FOR_JAVA" \
+  -Dpython.executable="$PRG_FOR_JAVA" org.python.util.jython $JYTHON_OPTS "$@"
 JYTHON_STATUS=$?
 
 if [ -n "$profile_requested" ] ; then
-  echo "Profiling results:"
-  cat profile.txt
+  if [ -f profile.txt ]; then
```
Stewori: Interesting. Does it really happen that profiling is requested but no profile.txt is written? Wouldn't that case justify an error message on its own right?
Is this check really Java 25(or 21) compatibility relevant or is it an independent improvement?
That would not be an issue, btw. We usually encourage to make minor improvements as side-edits of actual fixes. But this PR is so large that I ask these questions to sort out a bit what's going on.

src/shell/jython.py
```diff
@@ -247,6 +251,58 @@ def jython_home(self):
         self._jython_home = home
         return self._jython_home
 
+    def needs_jdb_path_alias(self):
```
Stewori, line 254: Perhaps this would be a good place to explain why a path alias may be needed. What's the issue with jdb being fixed here (and in jython shell file above). A doc comment explaining the reasons behind this in some detail would be good.

NEWS
```diff
@@ -16,6 +16,10 @@ Jython 2.7.5
 
 New Features
 
+    - Java 25 is supported when running Jython built for its lowest supported
+      Java bytecode level. The launchers add the Java 25 native-access options
+      needed by Jython's native dependencies.
+
```
Stewori, line 22: Thanks for maintaining the NEWS file!

ohumbel:
@robertpatrick Then I am sorry to suffer under the "it's not working on my machine" syndrome.

The pipelines decide.

Therefore I won't bother you with failing Ubuntu and Windows, both behind corporate firewall and proxy.

Lib/test/test_socket.py
robertpatrick: It allows the tests to pass in an environment where this hardcoded IP address is actually assigned to a machine in the network.  Hardcoding an IP address and assuming it will be unreachable in the test was naive…

Lib/test/test_httplib.py
robertpatrick: I checked this, but `www.pythontest.net` is not a suitable replacement for this specific test because it intentionally presents a self-signed certificate. This test verifies that default HTTPS validation succeeds for a publicly trusted certificate, so it still needs a host like `www.python.org`. I added a comment to make that  distinction explicit.

src/org/python/util/jython.java
robertpatrick: Just silencing noise

Lib/test/test_urllib2.py
robertpatrick: Good point. The `_opener` backup was unnecessary for this change. The issue is only that `urllib2.urlopen()` lazily builds the global opener from the current proxy  environment, so the test should clear proxy-related env vars before running doctests/unit tests. I removed the `_opener` save/restore and kept the environment guard only.

I will update the PR to remove this.

Lib/test/test_urllibnet.py
robertpatrick: Yes, that is the reason. `TEST_HTTP_URL` is `http://www.pythontest.net`; these urllib tests need a plain HTTP endpoint, while `www.python.org` now redirects to HTTPS  and changes the behavior being asserted. I added a short comment to make that explicit.

I updated the PR to add a comment on this.

Lib/test/test_urllibnet.py
robertpatrick: Yes, it is still `test_support.TEST_HTTP_URL`, which is the pythontest.net root document. `urlretrieve()` stores that response body in a local temporary file; this  test only checks that response headers, especially `Date`, are available. I added a comment to clarify that the URL does not need to point to an image file.

src/shell/jython
robertpatrick: Good catch. That guard was defensive from the profiling investigation, but it is not needed in the final version. Java 13+ now exits before launching the bundled profiler, and supported Java versions should produce `profile.txt` after the `profile.properties` fix. I restored the unconditional `cat profile.txt` so missing
  profiler output remains visible.

robertpatrick:
> @robertpatrick Then I am sorry to suffer under the "it's not working on my machine" syndrome.
> 
> The pipelines decide.
> 
> Therefore I won't bother you with failing Ubuntu and Windows, both behind corporate firewall and proxy.

No worries, I am also behind a  corporate firewall and proxy.  I was able to get it to pass in that environment by exporting the JAVA_OPTS environment variable with the necessary proxy -Ds.

src/shell/jython.py
robertpatrick: Agreed. I added comments in both launchers explaining the workaround. The issue is that some `jdb` implementations, notably OpenJDK/Temurin 8, split the debuggee  command line on spaces when launching the second-stage JVM. The temporary no-space symlink keeps the classpath and `-Dpython.*` path values intact when Jython is installed under a path containing spaces.

Lib/test/test_urllibnet.py
robertpatrick: This is a real change, and I checked whether we could restore the HTTPS URL. With an HTTP proxy configured, legacy `urllib` reports `geturl()` as `http://www.python.org/` for an `https://www.python.org/` request, so the old HTTPS variant fails in proxied environments. This test is only asserting that `geturl()` returns the opened URL, not HTTPS behavior, so I kept it on the shared HTTP test URL and added a comment explaining why.

Stewori:
For the record, as always, when binary files occur in a PR I routinely verify their authenticity. I hereby confirm that the provided binary files in this PR `asm-9.9.1.jar`, `asm-util-9.9.1.jar` and `asm-commons-9.9.1.jar` are binary identical to those provided on maven central.

@robertpatrick Thanks for addressing all of my remarks thoroughly!

lib-python/2.7/test/test_mailbox.py
jeff5, line 1: `lib-python/2.7` contains files taken unmodified from CPython, and that may be refreshed from there. (Long ago, that refresh was automatic.) We do not modify them.

Strategy 1 in response to a bug is to look upstream to CPython 2.7.x (perhaps the bug has been fixed there), and take that plus any other changed files it depends on. When necessary (Strategy 2), we make a copy in `Lib` that will shadow the CPython one during the build, but obviously that disables future use of Strategy 1 and complicates subsequent maintenance.

If it's just a convenient improvement, rather than essential, we'd probably accept the inconvenience. Startegy 2 is quite common for tests, however.

lib-python/2.7/test/test_smtpnet.py
jeff5, line 1: `lib-python/2.7` contains files taken unmodified ...

lib-python/2.7/httplib.py
jeff5, line 1: `lib-python/2.7` contains files taken unmodified ...

Lib/test/test_float_jy.py
```diff
@@ -21,10 +21,16 @@ def test_float_repr(self):
 
     def test_float_repr2(self):
         # Quite possibly these divergences result from JDK bug JDK-4511638:
+        if jython and test_support.get_java_version() >= (21,):
```
jeff5, line 24: Seems like my speculation here was correct (JDK-4511638 claims fixed in version 19).

src/org/python/util/jython.java
jeff5: I think that in normal circumstances, a user or application manager might want to know that they aren't going to get properly formatted log messages. ISTR it emerges as "noise" in a test involvig a security manager, but I've never found it objectionable.

build.gradle
```diff
@@ -650,7 +670,9 @@ jar {
         attributes ([
             'Main-Class': 'org.python.util.jython',
             'Built-By': 'build.gradle',
-            'Automatic-Module-Name': 'org.python.jython2'
+            'Automatic-Module-Name': 'org.python.jython2',
+            'Enable-Native-Access': 'ALL-UNNAMED',
```
jeff5, line 674: There are integrity reasons why recent JVMs make us jump through these hoops. Does this weaken integrity for the application as a whole that consumes this JAR, to have it ? I think it is fair to do this for the standalone JAR and for the launcher as they are meant to be packaged solutions for the user relatively unsophisticated in Java, to run Python in a Java environment.

Would someone constructing a Java application, consuming the slim or fat JAR, rather have the warning and then control this themselves (with help from our documentation, that is)? Genuinely an open question to a more expert group than myself.

src/shell/jython.py
jeff5, line 1: A manual step is necessary to turn this into jython.exe. I think this is why I still get test failures on Windows.

We can leave that to a separate issue, I think. Quite a lot of binary blobs in this PR already.

If it works when launched using CPython and this script on Windows, then it (very probably) works.

src/shell/jython
jeff5, line 1: Nice to see this get some attention as I'm not skilled in shell script so it falls behind `jython.py`.

jeff5 (changes requested):
@robertpatrick  A lot of relly good stuff here as commented earlier.

The changes made to permit testing in a proxied environment are a worthy objective in their own right, independent of the Java 25 compatibility. They're not things I know how to test but you're obviously a sophisticated user. (We'll credit you not Codex.)

The only blocker I think is that the PR must not modify `lib-python`. An alternative exists.

Other comments are either appreciative or a question.

src/shell/jython.py
robertpatrick: I am still testing on Windows and trying to work through the issues.

build.gradle
robertpatrick: this is the standard way to enable native access in JDK25 so we can either accept that or go chasing wild geese... :-)

As a consumer of a JAR, I would expect that the JAR works and loads/executes what it is supposed to do rather than being crippled and forcing me to make a decision about the internals of the JAR without knowing all of the implications on this dependent JAR if I choose not to allow the native code to load.  As such, I think that we need to do this rather than breaking encapsulation and forcing decisions on users who don't have, and arguably shouldn't have to have, internals knowledge of the dependent JAR and make decisions aboutit.

lib-python/2.7/httplib.py
robertpatrick: ok, but what is the alternative?  That only "special" people (i.e., those who can operate outside of a corporate firewall/proxy environment) can build and test Jython?  It seems that I am not the only contributor that is facing such problems.  I mean, we could let the tests fail but then it is difficult to fully test PRs since those tests will always fail in such an environment.  I am using AI to help with this PR--doesn't everyone these days?  In order to not have to pay for the millions of tokens consumed in this adventure, I must be in my corporate environment.

Would you prefer a big top-level switch that simply skips all of these these problematic tests (e.g., the tests written without the ability to route through a proxy or ones that rely on httplib.py to make calls to external web sites) in a corporate environment?

The other point to consider is what is the lifecycle status of CPython 2.x.   It is my understanding that CPython 2.x is no longer supported/maintained.  Regardless, how likely is it that Jython is going to pick up a newer version of this CPython 2.7.x library?

src/org/python/util/jython.java
Stewori: I've been pondering this noise-silencing decision, too. IMHO noise silencing is not necessarily a bad thing. Ideally, the warning would pop up only in verbose mode. Would `Level.FINE` achieve this? Since this tiny detail has not the urgency of Java 25 compatibility, I would ideally like to see the discussion decoupled and in a separate issue. However, I'm personally fine with keeping the change for now as-is. We can open a separate issue for this and assess if it can be made somehow verbose-mode only and adjust it in a new PR.

lib-python/2.7/httplib.py
Stewori: Having an overwrite-copy of the file in the Lib folder is merely a formality. Just copy your version there and revert the lib-python to its original. No big deal, really. Breaking the rules now may cause confusion in the future and elevate maintenance burden. Changing the rules is not worth the effort either. It's best to just do the copy and revert.

robertpatrick: Even if we dream about support for Python 3 in Jython, a huge barrier for migration exists.  Our customers have millions of existing scripts written to Jython 2.2.1 and Jython 2.7.x standards.  Breaking customer scripts is a non-starter for us so before we could ever consider adopting this mythical future version of Jython, we would need to solve that problem.

Lib/test/test_float_jy.py
robertpatrick: would you like me to change this line to apply this to 19 and above (untested)?

src/shell/jython.py
robertpatrick: I am working through the test issues on Windows.  One big issue that I do not plan to address is that the native jython.exe requires extra args on JDK 11+.  Codex says:

``` 
I found the problem: the new test assumed every launcher auto-adds the Java 9/25 options. That is true for the script launchers (src/shell/jython and jython.py), but the Windows native jython.exe only forwards what it is given, mainly through JAVA_OPTS.

I’m going to fix this in two places: make Ant pass the built-in JDK 9/25 test JVM options through JAVA_OPTS for  launcher-based exec tests, and make the launcher test distinguish script-launcher auto-add behavior from Windows native launcher forwarding behavior.
```

My recommendation is a separate PR to address this to eliminate the need for the `JAVA_OPTS` to make the native launcher work properly with the newer JDKs.

robertpatrick: I created #426 to track this issue.

robertpatrick:
@jeff5 @Stewori I just updated the commit with the changes made for Windows build and test (except for Issue #426) and moved changes out of `lib-python/2.7`, as requested.  Since the PR is large, I will give you Codex's enumeration of the changes, which will hopefully make it easier to review:

- `build.xml`
    - Added `jython.launcher.test.java.opts`, combining `locale opts`, caller-supplied `jython.test.java.opts`, JDK 9 module opens, and JDK 25 native-access/Unsafe opts.
    - Changed launcher-based test targets to pass that through `JAVA_OPTS`.
    - Added `JAVA_OPTS` to the Windows launcher invocations that previously did not pass it.
    - Kept the fix scoped to build/test execution, since the Windows native launcher issue #426 is now tracked separately.
- `Lib/test/test_jython_launcher.py`
    - Uses a forward-slash synthetic Windows classpath path for `--print` parsing tests.
    - Added coverage for modern Java launcher options.
    - On Windows, verifies the native launcher forwards `JAVA_OPTS`.
    - On Unix-like launchers, verifies scripts synthesize JDK-version-specific args.
    - Fixed Windows detection so Git/MSYS/Cygwin `uname` on PATH does not make the test treat the native Windows launcher like a Unix script launcher.
- `Lib/test/test_support.py`
    - Updated `EnvironmentVarGuard` bookkeeping for Windows case-insensitive environment variables.
    - It now canonicalizes only the internal saved-state key, while preserving the caller’s original spelling foractual `os.environ` operations.
    - This prevents `http_proxy` / `HTTP_PROXY`-style aliases from being restored and then accidentally deleted during cleanup.
- `Lib/test/test_os_jy.py`
    - Added a Windows-only regression test for `EnvironmentVarGuard`.
    - The test verifies that unsetting both lowercase and uppercase spellings restores the original value correctly.
- `Lib/test/test_socket.py`
    - Added UDP client socket cleanup in `ThreadedUDPSocketTest.clientTearDown()`.
    - The comment documents the Windows/Netty hang seen in full regrtest runs, where delayed datagram channel cleanup  could leave later threaded UDP tests blocked.
- `Lib/test/test_urllib2_localnet.py`
    - Added as a `Lib/test` overlay copy instead of modifying `lib-python/2.7/test`.
    - Only intentional change from upstream is a `tearDown()` that resets the process-global urllib2 opener with  `urllib2.install_opener(None)`.
    - This prevents the localhost test’s proxy-disabled opener from leaking into later urllib2 network tests that need the real proxy environment.
- `tests/java/javatests/Issue2455Test.java`
    - Changed the test to append the temporary directory directly to `sys.path` through `PySystemState`.
    - Avoids embedding a Windows path with backslashes into a Python string, where backslashes can be interpreted as escapes.
- `tests/java/org/python/indexer/TestBase.java`
    - Switched source loading to `Util.readFile(path)` so raw line endings are preserved.
    - Adjusted the `testGetSource()` assertion to normalize CRLF only for that assertion.
    - This keeps indexer offsets consistent on Windows checkouts.
- `tests/modjy/build.xml`
    - Added a default empty `jython.test.java.opts`.
    - Passes `jython.test.java.opts` into the forked `modjy` Java test process.
    - This lets matrix runs propagate shared proxy/socket JVM properties into `modjy` tests.

robertpatrick:
I just pushed what I hope to be the last cleanup changes to the PR.  One of the things in this commit is trying to clean up race conditions between the socket tests, which were mostly relying on implicit closing of sockets bound to ports that are reused across mutliple tests, primarily in test_socket.py.  Note that my Windows VM did not see these errors but my MBP M4 Pro did.  I do not believe that this PR introduced these race conditions and I am hopeful we have cleaned at least some of them up.  However, race conditions are difficult to detect so I honestly cannot say that the PR fixes every one of them.

Lib/test/test_float_jy.py
jeff5: No. I think if it works at LTS versions, that's enough. If Java 19 becomes an issue the JDK bug reference is enough of a clue.

lib-python/2.7/httplib.py
jeff5: The current `lib-python/2.7` is a snapshot at around 2.7.10, with some careful updates to treat specific bugs. Wholesale update to 2.7.18 (the last CPython 2) is a challenging objective, but that or further piecemeal updates are real possibilties.

I'm glad you found the shadowing convention possible to use (Strategy 2 in my first note).

ohumbel:
I configured a matrix to build with java 8 and then run the ci tests with java versions up to 25.
GitHub actions do not offer java 8 on every platform any more, therefore the test runs start with 11.
The `test_socket*` tests turned out to be quite flaky, so I excluded them.
The results can be found on this [local pull request](https://github.com/Cafino/jython/pull/1).

If there are changes to PR #424, I can manually trigger the pipelines again. 

Update 2026-05-19: The link is now working

src/org/python/util/jython.java
jeff5: Supporting Java 25 does not motivate a change from WARNING to FINE. By all means take the discussion elsewhere. I have learned not to change behaviour without a positive reason.

robertpatrick:
> @robertpatrick A lot of relly good stuff here as commented earlier.
> 
>  I think that in normal circumstances, a user or application manager might want to know that they aren't going to get properly formatted log messages. ISTR it emerges as "noise" in a test involvig a security manager, but I've never found it objectionable.

@jeff5 I think the problem is that this line is doing two different jobs at the moment. It reports both:

- “we couldn’t set the preferred SimpleFormatter format”, and 
- “we couldn’t install/configure the org.python console handler”. 

The first is mostly a formatting degradation and can be noisy under a security manager; the second is more operationally significant because -v logging may not behave as intended.

My suggestion is to split those cases instead of using one final message for both. I’d log the formatter-property failure at FINE or CONFIG, but keep a WARNING when we fail to attach/configure the console handler. That preserves the useful signal for users/application managers without making the security-manager test noise drive the level for both cases.

src/org/python/util/jython.java
robertpatrick: @jeff5 Well...the fact is that this logger serves two purposes so Codex was simply trying to clean up unnecessary noise in the test.  We could either split these two purposes so that each logs a distinct message at the right level or just restore it to `WARNING`.  At the end of the day, I could care less what we do here.

However, let's not pretend that there isn't technical debt in the code...

jeff5:
I think that's a good analysis. Also, the change is not of the essence for Java 25.

ohumbel:
> I configured a matrix to build with java 8 and then run the ci tests with java versions up to 25. GitHub actions do not offer java 8 on every platform any more, therefore the test runs start with 11. The `test_socket*` tests turned out to be quite flaky, so I excluded them. The results can be found on this [local pull request](https://github.com/Cafino/jython/pull/1).

There was a typo in the link - sorry for that.
It is now correct: [https://github.com/Cafino/jython/pull/1](https://github.com/Cafino/jython/pull/1).

src/org/python/util/jython.java
jeff5: This is a good observation on this bit of code, but has nothing to do with Java 25. I am content with this noise.

tests/java/org/python/indexer/TestBase.java
jeff5, line 1: Did you encounter a problem with line-endings? This test runs fine for me on Windows. The test material has LF line-endings and we use `autocrlf=input`.

tests/java/javatests/Issue2455Test.java
jeff5, line 1: This test runs fine for me., but I see now that I have just been lucky with backslashes and Python 2 is tolerant about them.

So it's a bug found, but it has nothing to do with Java 25, so thank you Codex and I shall extract this one to a separate change set. I have a smaller fix anyway.

Is this Codex being an enthusiastic novice? Noticing unrelated bugs and being unable to resist the temptation to fix them.

Lib/test/test_mailbox.py
```diff
@@ -0,0 +1,2124 @@
...
+        self._box.remove_folder('two')
+        self.assertEqual(len(self._box.list_folders()), 0)
+        self.assertEqual(self._box.list_folders(), [])
+
+    def test_clean(self):
+        # Remove old files from 'tmp'
+        foo_path = os.path.join(self._path, 'tmp', 'foo')
+        bar_path = os.path.join(self._path, 'tmp', 'bar')
+        self._delete_recursively(foo_path)
+        self._delete_recursively(bar_path)
```
jeff5, line 636: Only these two lines are new. They don't exist in the latest upstream test. (Other improvements do.)

Again, this seems to be Codex making improvements unrelated to the objective. The original runs fine for me. Did you really have trouble? If not, I'd revert the shadowing file.

Lib/httplib.py
```diff
@@ -0,0 +1,1473 @@
...
+                thelen = str(len(body))
+            except (TypeError, AttributeError):
+                # If this is a file-like object, try to
+                # fstat its file descriptor
+                try:
+                    thelen = str(os.fstat(body.fileno()).st_size)
+                except (AttributeError, OSError):
+                    try:
+                        thelen = str(os.path.getsize(body.name))
+                    except (AttributeError, OSError, TypeError):
```
jeff5, line 1103: These 3 lines are the reason for creating a shadow copy of the file. But why are they here? CPython 2 does not find it necessary to add this. (Things are changed a lot in Python 3, so I can't tell.)

I would revert if there is no compelling reason.

Lib/test/test_urllib2_localnet.py
```diff
@@ -0,0 +1,556 @@
...
+        super(TestUrlopen, self).setUp()
+
+    def tearDown(self):
+        # setUp installs a process-global opener with proxies disabled so these
+        # localhost tests ignore any external proxy environment. Restore the
+        # default opener afterwards: on Windows matrix runs the next urllib2
+        # network tests rely on the preserved proxy environment to reach
+        # external HTTP resources.
+        urllib2.install_opener(None)
+        super(TestUrlopen, self).tearDown()
```
jeff5, line 364: So the motivation is the proxied environment, even though this test does not use external network resources, because use of `install_opener` in `setUp()` messes up subsequent tests that do. And the change entails the shadow copy. :(

I think CPython is treating the same problem here: https://github.com/python/cpython/blob/8d21aa21f2cbc6d50aab3f420bb23be1d081dac4/Lib/test/test_urllib2_localnet.py#L470 . Might this be a better solution?

I think I would like longer term to update the `lib-python` version from CPython and remove the shadow. However, it has a lot of other differences: useful differences but with unknown impact.

jeff5 (commented):
Having studied this carefully, I now understand it in three main parts:
1. Bringing JARs up to date.
2. Compatibility with Java 25: mainly native access for Java 25, but includes `test_float_jy`.
3. Changes to network support to permit testing behind a proxy.

These are all useful. There are other minor tweaks in the PR too, but I have commented where I think these are misconceived or unnecessary. I'm open to correction on those comments.

I would like to split the PR into those parts so that contributors coming after may more easily associate a change to its intent. They pretty much split by file, although `build.xml` has elements of them all.

As you've said @robertpatrick that you don't have the time for that reorganisation, I'll do it and aim to credit you in the commits. If some end up with my name on, it's my lack of `git` foo, not intentional.

tests/java/org/python/indexer/TestBase.java
robertpatrick: @jeff5 things are not as clear cut as relevant to JDK 25 or not.  Many tests don’t clean up after themselves and create race conditions.  Since the ONLY way to validate changes are the tests, I had no choice but to fix issues causing test failures.  Since many of the test changes were addressing race conditions in the tests, by definition, slower machines may not observe these test failures while faster machines will.  Do whatever you want but it is naive to think that just because you don’t observe an issue yourself that the issue doesn’t exist.

I spent many hours (and hundreds of millions of tokens) creating and testing the changes in the PR across Mac, Linux, and Windows.  How many other contributors do this?

Lib/test/test_mailbox.py
robertpatrick: Yes, more race conditions

Lib/httplib.py
robertpatrick: Whether or not some random, CPython version from 20 years ago felt the need to add this, I was seeing test failures on at least one platform due to this.  Again, like caused by race conditions.  If the tests don’t pass on modern machines, how do you plan to validate the changes?

tests/java/org/python/indexer/TestBase.java
jeff5: The particular comment is about whether this test encounters CRLF line endings, which it is hard to see as a concurrency, a processor speed or a clean-up issue. If you see such line endings in the files themselves, they were most likely added by `git` during checkout.

Taking your comment as general to the review, yes, some tests can be flakey, and none of them are designed to run concurrently: many use the same file name as a temporary for example. I also understand that you would want to make the test suite run in your proxied environment before working on the Java 25 specific fixes.

robertpatrick: The race conditions have nothing to do with running tests concurrently.  They are all about proper resource cleanup (e.g., ports, mailboxes, etc.) when one test completes so that the next test being run can reuse the same resource without encountering unexpected conditions (e.g., port already in use, mailbox already exists, etc.).  

One common pattern I found was that the tests were relying on JVM garbage collection to clean up these external resources.  That approach is not reliable and creates a race condition on whether or not GC clears the resource prior to the next test trying to use it.

Do you have an ETA when the changes will be merged?

jeff5: > One common pattern I found was that the tests were relying on JVM garbage collection to clean up these external resources. That approach is not reliable and creates a race condition on whether or not GC clears the resource prior to the next test trying to use it.

Yes, we've fixed tests ourselves that (rather) rely on CPython garbage collection, which is immediate, of course. Guidance says you shouldn't rely on that but authors are not reminded by failure.

It's reasonable to think I can get it ready to merge over the weekend.

jeff5:
Just noting here that asm JARs are already at version 9.10.1 on Maven Central so I'll do an independent PR for that, instead of part 1 of this.

Lib/httplib.py
jeff5: It turns out the root cause is that `os.fstat` (line 1099) fails on later Java versions unless `--add-opens=java.base/java.io=ALL-UNNAMED` is given. But since other parts of the patch add that, I think we're safe. I'll be able to test that.

Lib/test/test_jython_launcher.py
```diff
@@ -13,7 +13,13 @@
 launcher = None
 uname = None
 is_windows = False
-some_jar = os.path.join(os.sep, "a", "b", "c", "some.jar")
+# Use forward slashes on Windows because these tests parse launcher --print
```
jeff5, line 16: I don't find this to be a problem. (If it had been, it would be a problem on already supported versions of Java.)

Lib/httplib.py
jeff5: I tested that, and we're safe.

jeff5:
I think this is superseded by #435 and #437 (and a few supporting acts). Thhgis has been the source of multiple useful contributions, expressed in those PRs, so thank you @robertpatrick .

PR #427, opened 2026-05-17
Fix the Maven badge for Jython in README.md

wfouche:
image

fixed and now displayed as

image


wfouche:
@Stewori, consider merging this one as well given how simple it is.

PR #428, opened 2026-05-19
Use jffi version 1.3.15

wfouche:
Upgrading jffi from 1.3.14 to 1.3.15 (with improved support for newer versions of Java).


Stewori:
I hereby confirm authenticity of the attached binary - it's the same file as is on Maven Central.

Stewori:
I noticed that an earlier update to jffi 1.3.14 is advertised in NEWS under "New Features" for 2.7.5a1. Please adjust that line to 1.3.15.

wfouche:
> Please adjust that line to 1.3.15.

Done.

Stewori:
@jeff5 since this is a trivial change, I'd like to merge it unburocratically. Now since we got people to maintain NEWS I notice with my own PR that this is prone to create merge conflicts for the NEWs file all the time. These are trivial to resolve, but still cause some extra work every time. Probably we don't need to deal with this specially since it's low volume. (But it's part of the reason I'd like to merge simple PRs like this one rather sooner than later.)

jeff5:
Although JAR updates are better left until the beta is close, to avoid duplication/bloat. It is close, so that's fine.

Issue #416, opened 2026-03-04
Assign explicit serialVersionID to PyObject and subclasses.


Stewori:
Triggered by https://github.com/jython/jython/issues/234.
For better or worse, `PyObject` is `Serializable`, but serializations are unnecessarily incompatible accross Jython versions because there are no explicit serialVersionID fields. (See https://apidia.net/java/OpenJDK/25/?pck=java.io&cls=.Serializable for the background of this field)
This should be improved to make the usecase from #234 more stable, but also on its own right.
Automatic serialVersionIDs may change due to method reordering or all sorts of modifications that do not actually introduce a serialization incompatibility. Still, in those cases object deserialization is broken.  When choosing version values, we can assign them such that compatibility to *one* previous Jython release is retained.
By the method explained in https://stackoverflow.com/questions/22169268/is-it-possible-to-get-serialversionuid-of-the-java-class-in-runtime it is possible to retrieve e.g. the ids for Jython 2.7.4 and then pin them for future versions in 2.7.5 onward.
What makes this a bit tedious is that it's not enough to do this for PyObject, but it has to be done for every subclass on its own. (Some script will get all version ids automatically.)
Still, I think it's crucial to fix this ASAP (i.e. for 2.7.5) to avoid further divergence.
Since 2.7.4 there have not been many changes, so pinning the versions should work. We can check for which classes versions do actually differ and examine those cases more closely.

Still have to think about derived classes, since they are generated. Perhaps it's best to let them be. We would have to touch the generator in a non-trivial way. Since they haven't been re-generated in a long while, maybe their serial versions did not even change for ages. (Unless changes in the parent class trigger a version change for subclasses.)

It would already be a win to pin the versions for the main builtin types like PyString, PyInt, PyList, ... since they are used most heavily. They are also the typical data types for serialization.

Ideas? Opinions?

yogi1967:
Thanks for raising this issue. As the ‘author’ of #234 I fully support doing this. I realise this doesn’t help with the technical analysis of the proposal, but it certainly seems sensible to pin the serial ids in this way. I would even suggest that they are pinned for more than just one version and remain stable unless something has actually changed? If this does go ahead could we grab the generated IDs from 2.7.2 and reuse those?

Stewori:
[...]

Issue #234, opened 2023-02-05
Clarification on compile 


yogi1967:
Apologies this isn't as such an issue report, but seeking clarification: Jython 2.7.x

For reasons outside of my control, I am forced to use single, large script files.. I write extensions to an app which controls the environment my code executes in (code is started / run by the master app through PythonInterpreter). 

I cannot do 'normal' imports as my code runs from within a ZIP file. I can get the zip resources and contents via classloader.getResourceAsStream() etc.

Because I have large files, I sometimes hit the:
java.lang.RuntimeException: Module or method too large in xxx.py

and it then says:

Please provide a CPython 2.7 bytecode file (.pyc), e.g. run
python -m py_compile xxx.py

Alternatively, specify a CPython 2.7 command via the python.cpython2 property, e.g.:
    jython -Dpython.cpython2=python
or (e.g. for pip) through the environment variable JYTHON_OPTS:

----
So, question 1. Why is it saying use python -m for jython?

question 2: is it actually possible to create .pyc files for jython, and if so, how do you actually do it?

question 3: what is the difference between .pyc files and xxx$py.class files?

I have now managed to compile jython using: 
java -cp jython.jar org.python.util.jython -c "import compileall; compileall.compile_file('xx.py')"
.. and run it using imp.load_module() so this bit is great! But of course the script ends up within a sub module scope whereas I really want it within the scope of the original script. (Hence execfile would be  more useful). 

question 4: is there a way to execute $py.class files using execfile() via a resource stream (like load_module) rather than from a file?

question 5: I see exec can run Code objects.... What is this Code object? Is it the same as a .pyc file and/or $py.class file? And again, can a stream be executed rather than a file....?

Separately, I have seen that execfile() on a script which hits the too large issue, does actually run if a .pyc is also present on disk in the same location and one that was compiled by python -m py_compile... So something with pyc works... This all seems very odd..!?

Sorry for the questions, I have done much reading and experimenting. I am glad I have finally got something working, but your clarification(s) would be appreciated? 

Thanks

Stewori:
On short time, so just the essentials:
Jython *can* read and execute CPython's (2.7) bytecode pyc files, but it cannot generate them. $py.class files is Jython's original way of bytecode, i.e. compile to Java bytecode. Support of pyc files is just an additional feature. Jython had frequent issues due to limited method length in Java's bytecode spec. Using the pyc file support is a workaround for that.

yogi1967:
[...]

PR #429, opened 2026-05-26
Improve backward compatibility of oversized function handling

Stewori:
Addresses #416 and #234.


jeff5:
I've pulled this so I can test it on Windows and Java 25. No problems so far. (Also, this test is fine: https://github.com/jeff5/jython/actions/runs/29184713052/job/86628610171 . Since we have to install 2.7 these days, I was worried it might not find it.)

Stewori:
As a side-effect, this PR eliminates the need for Python 2.7 on Windows if jython.exe is used as a launcher. I assume the launcher is built with PyInstaller via Python 2.7. Now, jython.exe itself can be used in the cpython2 property (and is set so by default). Note that I did not rebuild jython.exe since I'm not using windows. So I kindly ask you to handle this bit.

jeff5:
> As a side-effect, this PR eliminates the need for Python 2.7 on Windows if jython.exe is used as a launcher.

Oh that's clever! Yes, I'm just getting to that bit of the code. I felt I should read it, but TBH, I don't expect to find anything to object to.

jeff5:
LGTM. Very conscientiously written. In the Java 25 and 17 regrtests that I run ad hoc, it only stumbles on test_socket, which we always do. https://github.com/jeff5/jython/actions/runs/29197785412/job/86663753617. I'll push the regenerated jython.exe to this PR, then merge.

Stewori:
Thx, Jeff! I had some AI help, but it was still a long process and dialog to recognize every remaining edge case and close every gap. I'm glad to get this PR finalized and hope, the oversized-method-mechanism is bullet-proof now for the future.

jeff5:
> I had some AI help

Oh, in that case, I think the `loadPyBytecode` is too long and rambling, but it's not a blocker. 

I find I can't push `jython.exe` to this, so I'll do it separately.

Issue #434, opened 2026-06-02
CVE-2023-34462(CVSS 6.5), CVE-2022-41915(CVSS 6.5) Jython 2.7.4 using a vulnerable Netty version 4.1.173.Final


GrigorStIvanov:
Jython bundles a bunch of external libs, many of which are rather old versions. The Netty used by the Jython dependency has two vulnerabilities related to it.
So, the entire Jython is red-flagged with CVSS 6.5 CVE vulnerability.

Please update to the latest version in 2.7.5.

Stewori:
Yes... updating Netty turned out not so easy #349. We pinned the old version for a reason. Yet another lesson how backwards-breaking changes are evil. Even in terms of security...
(no worries, I finally fixed #349, PR will come)

GrigorStIvanov:
Awesome, thanks for the fast feedback @Stewori.

jeff5:
Merged, so now at 4.1.136 (thanks @Stewori).

PR #442, opened 2026-07-20
Update Netty to 4.1.136.

### LINKED ISSUE #349: Updating Netty JARs leads to errors in `test_httplib` and others
I tried updating Netty to 4.1.112 (see #342), but it gives me errors in:
```
test_httplib
test_robotparser
test_smtpnet
test_ssl_jy
test_urllibnet
```
The error is always:
```
SSLError: [Errno 1] SslHandler removed before handshake completed
```
This suggests a race with the Netty asynchronous processing. Tear-down of Netty is known to be a fragile aspect of the tests.

The latest version known to work at this time is 4.1.73

_Originally posted by @jeff5 in https://github.com/jython/jython/issues/342#issuecomment-2243576808_

Stewori:
Fixes #349


Stewori:
The magic here is merely smoke and mirrors: Newer Netty was just missing `netty-transport-native-unix-common`. The error we observe in #349 is actually a disguised `java.lang.NoClassDefFoundError: io/netty/channel/unix/UnixChannel`. Netty is guilty of the anti pattern of catching also errors (not just exceptions) at some critical point, thereby swallowing the `NoClassDefFoundError` and replacing it with some misleading other error. Mission accomplished.
Note that there exist platform-specific versions of the artifact. For our purpose, however, the generic one should be sufficient.

Stewori:
Ah, this also fixes #434 , forgot to mention it in NEWS. However, rerunning tests just for this little entry is not worth it. I'll add it with the next chunk of dependency updates.

jeff5:
> Newer Netty was just missing `netty-transport-native-unix-common`. The error we observe in #349 is actually a disguised `java.lang.NoClassDefFoundError: io/netty/channel/unix/UnixChannel`.

Good detective work! Also, this will have been difficult to spot on my Windows box ... but is there a Windows equivalent I wonder? I may be able to check.

However, your PR is signalling that it cannot be merged. Somewhere I think you submitted one or more PRs from `master` and we squash-merged them, which invalidates your fork history. You may be able to recover by renaming `master` to something else, but I'm not sure how you create a new, correct one.

The bigger reset would be to rename your fork repo to something else, and make a fresh clone under the `jython` name, from the project repo. I think a lot of external references (like on your dev box) then recover because they are URLs to https://github.com/Stewori/jython, although on GitHub they get amended to follow the original (so it doesn't fix open PRs).

Stewori:
For me, it signals "Ready to merge". I would just hit "squash and merge" and be done. Including merges from master is just usual upstream sync. Github now does strange things when merge conflicts are fixed - it insists to merge into cloned master (or create yet another branch), which causes some additional noise. One time it opened on own behalf an upstream PR to Jython master, which I immediately closed again. Still, there is no indication for me that squash and merge would not hide away all this noise. For project history it should result just in one "Update Netty" PR.
Re Windows: It appears to me that for some reason also Windows uses some classes from transport-native-unix-common (perhaps just for interfacing). There does not seem to be a Windows equivalent. Yet, the error also occurs on Windows and regrtests (in the CI) pass on Windows with the fix included (or don't they test network features?)
Since I'm not testing on Windows locally, you might want to clone the fork and test on Windows before we merge.

jeff5:
Runs ok on my machine despite the tortured history and no windows-specificity.
Ah, ok squash-merge will be allowed. Maybe the repo hygiene can wait for a quiet moment.

jeff5:
> For me, it signals "Ready to merge". I would just hit "squash and merge" and be done.

The button (defaulted to rebase) was so greyed-out it looked like I was stuck, but actually one can still switch to squash and then it's green.

> Re Windows: It appears to me that for some reason also Windows uses some classes from transport-native-unix-common (perhaps just for interfacing).

That's odd but there we are.

> There does not seem to be a Windows equivalent. Yet, the error also occurs on Windows and regrtests (in the CI) pass on Windows with the fix included (or don't they test network features?)

They test with the network resource enabled. Some specific tests are excluded as troublesome, choices we ought to reconsider sometime.

> Since I'm not testing on Windows locally, you might want to clone the fork and test on Windows before we merge.

When I tested, I pulled down the PR branch with `git fetch upstream pull/442/head:update-netty-1`. The CPython devguide has a useful [cheat-sheet](https://jython-devguide.readthedocs.io/en/latest/gitbootcamp.html#downloading-other-s-patches).

jeff5:
With a squash, the only downside of the tortuous history in the end is that GH wants to include a lot of commmit messages from the apparent history.

PR #450, opened 2026-07-27
Update mysql connector to 9.7.0 and fix tests and references accordingly.

Stewori:


Stewori:
With this PR, all dependency updates are done. (Junit needs to remain pinned to 4.10 since we don't want to deal with hamcrest dependency). Running the database tests (stuff in Lib/test/zxjdbc) was a new experience for me and some of the related code looks like it hasn't been touched for well over a decade if not two. It requires setting up a local mysql process (easy e.g. via docker), for which we might write up some instructions in docs. But that's a task for after this release. With this PR tests for mysql connector pass again and I kept necessary fixes and adjustments as minimal as possible for now (me might revisit some of it later).

jeff5:
> Running the database tests (stuff in Lib/test/zxjdbc) was a new experience for me and some of the related code looks like it hasn't been touched for well over a decade if not two. It requires setting up a local mysql process (easy e.g. via docker), for which we might write up some instructions in docs.

Thanks for tracking down all the name changes. Yes, some notes would be useful. I think I dabbled in MySQL an age ago, before I started getting my kicks from Jython. I'll run some basic checks on my machine later today.

